Issue #8, opened 2015-09-11
Support for VCF v4.3


cyenyxe:
Support for VCF v4.3 changes and gVCF additions.

The following new checks must be implemented:
* SAMPLE meta entries structural changes
* PEDIGREE meta entries structural changes
* FILTER column: Duplicates not allowed

cyenyxe:
Closed via #37, although it is a known limitation that the "no duplicates allowed" check is not implemented yet (see #35).

PR #17, opened 2016-02-04
Support for multiple versions of VCF

cyenyxe:
Architecture for supporting multiple versions of the VCF specification (currently 4.1, 4.2 and 4.3). The CLI gets an argument _-v_ / _--version_ which internally allows to select the correct parsing state machine.

- Related to https://github.com/EBIvariation/vcf-validator/issues/8


src/vcf/main.cpp
```diff
@@ -58,34 +62,122 @@ namespace
             return 1;
```
jmmut: here this returns 1 if level is "stop": three lines above, it checks for "break"

src/vcf/main.cpp
```diff
@@ -58,34 +62,122 @@ namespace
...
         }
         
+        if (!vm.count("version")) {
+            std::cout << desc << std::endl;
+            std::cout << "Please choose one of the accepted VCF fileformat versions" << std::endl;
+	    return 1;
```
jmmut: indentation incoherence

PR #18, opened 2016-03-30
Multi-version testing architecture EVA-312

cyenyxe:
Architecture for testing against multiple versions of the specification. Previously, only one folder with input files existed, but these could never be valid against all versions because they require the first line (fileformat) to be different.

So for each version the following have been created:

* A cpp file for testing the parser (such as `test/vcf/parser_v3_test.cpp`)
* A folder with failed and passed input files (`test/input_files/v4.1/failed` and `test/input_files/v4.1/passed`)


inc/util/stream_utils.hpp
```diff
@@ -31,10 +31,12 @@ namespace ebi
...
         // `stream.get()` sets failbit as well on eof
-        do {
-            stream.get(c);
+        while (stream && stream.get(c)) {
             container.push_back(c);
-        } while (stream && c != '\n');
+            if (c == '\n') {
+                break;
+            }
+        }
```
jmmut: Did this fail?

cyenyxe: There were some issues when files ended with different number of newline characters. It was interpreting empty lines as another record, and the parser would go nuts.

PR #22, opened 2016-05-13
Error class hierarchy

jmmut:
This PR implements a error hierarchy to locate the errors programmaticaly, instead of leaving the location inside a string.

Highlights:
- `Error` is the parent class, and it inherits `std::runtime_error`.
- At the moment, an `Error` (or inheriting class) can receive a line number (from the input file), and a message.
- Every `Error`  (or inheriting class), has a default message, so only the line is mandatory.
- If a `Error` is `throw`n, it's advisable to `catch` it as a reference, to enable polymorphism.
- `ErrorPolicy` now only defines two methods: `handle_error` and `handle_warning`, so it is effectively decoupled from the location.
- `ParsingWarning` and `ParsingError` were removed. Use a combination of `Error` and `ErrorPolicy::handle_*`.
- Record and MetaEntry have their line number, in order to be able to construct an `Error`.
- Updated tests.

This PR allows for next steps:
- A new ErrorPolicy could be added that stores the errors for a future vcf_fixer.

### DIFF
--- a/inc/vcf/error.hpp
+++ b/inc/vcf/error.hpp
@@ -28,10 +28,13 @@ namespace ebi
     class Error : public std::runtime_error
     {
       public:
+        Error(size_t line) : Error(line, "Error, invalid file.") {}
+
         Error(size_t line, const std::string &message)
                 : runtime_error(std::string("Line ") + std::to_string(line) + ": " + message),
                   line(line),
                   message(message) {}
+
 
         // TODO: leave here or extract? if we will allow several levels of conservative-risky fixes
 //        virtual void fix(Error, ParsingState &state);
@@ -54,65 +57,94 @@ namespace ebi
     // inheritance siblings depending on file location
     class MetaSectionError : public Error
     {
+      public:
         using Error::Error;
+        MetaSectionError(size_t line) : MetaSectionError(line, "Error in meta-data section") { }
     };
 
     class HeaderSectionError : public Error
     {
+      public:
         using Error::Error;
+        HeaderSectionError(size_t line) : HeaderSectionError(line, "Error in header section") { }
     };
 
     class BodySectionError : public Error
     {
+      public:
         using Error::Error;
+        BodySectionError(size_t line) : BodySectionError(line, "Error in body section") { }
     };
 
     // inheritance siblings about detailed errors
     class FileformatError : public MetaSectionError
     {
+      public:
         using MetaSectionError::MetaSectionError;
+        FileformatError(size_t line) : FileformatError(line, "Error in file format section") { }
     };
 
     class ChromosomeBodyError : public BodySectionError
     {
+      public:
         using BodySectionError::BodySectionError;
+        ChromosomeBodyError(size_t line) : ChromosomeBodyError(line,
+            "Chromosome is not a string without colons or whitespaces, optionally wrapped with angle brackets (<>)") { }
     };
 
     class PositionBodyError : public BodySectionError
     {
+      public:
         using BodySectionError::BodySectionError;
+        PositionBodyError(size_t line) : PositionBodyError(line, "Position is not a positive number") { }
     };
     class IdBodyError : public BodySectionError
     {
+      public:
         using BodySectionError::BodySectionError;
+        IdBodyError(size_t line) : IdBodyError(line, "ID is not a single dot or a list of strings without semicolons or whitespaces") { }
     };
     class ReferenceAlleleBodyError : public BodySectionError
     {
+      public:
         using BodySectionError::BodySectionError;
+        ReferenceAlleleBodyError(size_t line) : ReferenceAlleleBodyError(line, "Reference is not a string of bases") { }
     };
     class AlternateAllelesBodyError : public BodySectionError
     {
+      public:
         using BodySectionError::BodySectionError;
+        AlternateAllelesBodyError(size_t line) : AlternateAllelesBodyError(line, "Alternate is not a single dot or a comma-separated list of bases") { }
     };
     class QualityBodyError : public BodySectionError
     {
+      public:
         using BodySectionError::BodySectionError;
+        QualityBodyError(size_t line) : QualityBodyError(line, "Quality is not a single dot or a positive number") { }
     };
     class FilterBodyError : public BodySectionError
     {
+      public:
         using BodySectionError::BodySectionError;
+        FilterBodyError(size_t line) : FilterBodyError(line, "Filter is not a single dot or a semicolon-separated list of strings") { }
     };
     class InfoBodyError : public BodySectionError
     {
+      public:
         using BodySectionError::BodySectionError;
+        InfoBodyError(size_t line) : InfoBodyError(line, "Error in info column, in body section") { }
     };
     class FormatBodyError : public BodySectionError
     {
+      public:
         using BodySectionError::BodySectionError;
+        FormatBodyError(size_t line) : FormatBodyError(line, "Format is not a colon-separated list of alphanumeric strings") { }
     };
     class SamplesBodyError : public BodySectionError
     {
+      public:
         using BodySectionError::BodySectionError;
+        SamplesBodyError(size_t line) : SamplesBodyError(line, "Error in samples columns, in body section") { }
     };
 
   }
--- a/inc/vcf/error.hpp
+++ b/inc/vcf/error.hpp
@@ -36,7 +36,7 @@ namespace ebi
                   message(message) {}
 
 
-        // TODO: leave here or extract? if we will allow several levels of conservative-risky fixes
+        // TODO: leave here or extract? if we will allow several levels of conservative-risky fixes, it's better to extract
 //        virtual void fix(Error, ParsingState &state);
 //        virtual void fix(MetaError, ParsingState &state);
 
--- a/inc/vcf/error.hpp
+++ b/inc/vcf/error.hpp
@@ -1,0 +1,106 @@
+/**
+ * Copyright 2016 EMBL - European Bioinformatics Institute
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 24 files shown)


inc/vcf/error.hpp
```diff
@@ -0,0 +1,153 @@
...
+  {
+    // parent of vcf errors
+    class Error : public std::runtime_error
+    {
+      public:
+        Error(size_t line) : Error(line, "Error, invalid file.") {}
+
+        Error(size_t line, const std::string &message)
+                : runtime_error(std::string("Line ") + std::to_string(line) + ": " + message),
+                  line(line),
```
cyenyxe: Better to use curly braces than brackets in constructors: https://stackoverflow.com/a/18222927 This applies both to `line` and `message`.

inc/vcf/file_structure.hpp
```diff
@@ -131,17 +138,18 @@ namespace ebi
...
-               std::vector<std::string> const & ids, 
-               std::string const & reference_allele, 
-               std::vector<std::string> const & alternate_alleles, 
-               float const quality, 
-               std::vector<std::string> const & filters, 
-               std::map<std::string, std::string> const & info, 
-               std::vector<std::string> const & format, 
-               std::vector<std::string> const & samples,
-               std::shared_ptr<Source> const & source);
+        Record(size_t const line,
```
cyenyxe: Is it necessary for this to be const? It's not in `MetaEntry`.

src/vcf/store_parse_policy.cpp
```diff
@@ -84,25 +84,22 @@ namespace ebi
...
+            auto key_values = std::map<std::string, std::string>{};
+            for (size_t i = 0; i < m_grouped_tokens.size(); i += 2) {
+                key_values[m_grouped_tokens[i]] = m_grouped_tokens[i+1];
             }
-        } catch (std::invalid_argument ex) {
-            throw ParsingError(ex.what());
+            state.add_meta(MetaEntry{state.n_lines, m_line_typeid, key_values});
+
+        } else {
+            // TODO Throw exception
```
cyenyxe: Now that we actually have those exceptions hierarchy, could a `MetaSectionError` be thrown?

src/vcf/store_parse_policy.cpp
```diff
@@ -161,37 +158,37 @@ namespace ebi
...
     void StoreParsePolicy::handle_body_line(ParsingState const & state) 
     {
+        // Transform the position token into a size_t
+        auto position = static_cast<size_t>(std::stoi(m_line_tokens["POS"][0]));
+
+        // Transform all the quality tokens into floating point numbers
+        float quality;
         try {
-            // Transform the position token into a size_t
-            auto position = static_cast<size_t>(std::stoi(m_line_tokens["POS"][0]));
```
cyenyxe: `std::stoi` could also raise an `std::invalid_argument` exception, although it should probably be in a different try-catch than the quality check.

src/vcf/meta_entry.cpp
```diff
@@ -23,21 +24,24 @@ namespace ebi
   namespace vcf
   {
   
-    MetaEntry::MetaEntry(std::string const & id)
-    : id{id}, structure{Structure::NoValue}
+    MetaEntry::MetaEntry(size_t line,
+                         std::string const & id)
+    : line(line), id{id}, structure{Structure::NoValue}
```
cyenyxe: Some more brackets that should be curly braces.

src/vcf/validate_optional_policy.cpp
```diff
@@ -146,7 +151,8 @@ namespace ebi
             state.add_well_defined_meta("contig", current_chromosome);
         } else {
             state.add_bad_defined_meta("contig", current_chromosome);
-            throw ParsingWarning("Chromosome/contig '" + current_chromosome + "' is not described in a 'contig' meta description");
+            throw MetaSectionError(state.n_lines,
+                    "Chromosome/contig '" + current_chromosome + "' is not described in a 'contig' meta description");
```
cyenyxe: If the exception in https://github.com/EBIvariation/vcf-validator/pull/22/files#diff-b7ed7bfe2cd51f747bc5f560a4647ee3R208 comes from checking a relationship between meta and body and throws a BodySectionError, shouldn't this one and the following in this file do the same?

src/vcf/vcf_v41.ragel
```diff
@@ -197,14 +197,18 @@
...
+            try {
+                // Check warnings (non-blocking errors but potential mistakes anyway, only makes sense if the last record parsed was correct)
+                OptionalPolicy::optional_check_body_entry(*this, ParsingState::records->back());
+            } catch (MetaSectionError &warn) {
+                ErrorPolicy::handle_warning(*this, warn);
+            } catch (BodySectionError &warn) {
+                ErrorPolicy::handle_warning(*this, warn);
+            }
+        } catch (BodySectionError &error) {
+            ErrorPolicy::handle_error(*this, error);
```
cyenyxe: There could be 2 try-catch blocks instead of 1 to avoid nesting, since `ErrorPolicy::handle_error` is only going to apply to `ParsePolicy::handle_body_line` anyway.

src/vcf/vcf_v42.ragel
```diff
@@ -194,12 +194,16 @@
...
-          ErrorPolicy::handle_body_section_error(*this, ex.what());
-        } catch (ParsingWarning ex) {
-          ErrorPolicy::handle_body_section_warning(*this, ex.what());
+            } catch (MetaSectionError &warn) {
+                ErrorPolicy::handle_warning(*this, warn);
+            } catch (BodySectionError &warn) {
+                ErrorPolicy::handle_warning(*this, warn);
+            }
+        } catch (BodySectionError &error) {
+            ErrorPolicy::handle_error(*this, error);
```
cyenyxe: There could be 2 try-catch blocks instead of 1 to avoid nesting, since ErrorPolicy::handle_error is only going to apply to ParsePolicy::handle_body_line anyway.

src/vcf/vcf_v43.ragel
```diff
@@ -194,12 +194,16 @@
...
-          ErrorPolicy::handle_body_section_error(*this, ex.what());
-        } catch (ParsingWarning ex) {
-          ErrorPolicy::handle_body_section_warning(*this, ex.what());
+            } catch (MetaSectionError &warn) {
+                ErrorPolicy::handle_warning(*this, warn);
+            } catch (BodySectionError &warn) {
+                ErrorPolicy::handle_warning(*this, warn);
+            }
+        } catch (BodySectionError &error) {
+            ErrorPolicy::handle_error(*this, error);
```
cyenyxe: There could be 2 try-catch blocks instead of 1 to avoid nesting, since ErrorPolicy::handle_error is only going to apply to ParsePolicy::handle_body_line anyway.

test/input_files/failed/failed_fileformat_000.vcf
```diff
@@ -1,4 +1,4 @@
-##fileformat=
+##qformat=
```
cyenyxe: An empty fileformat is already an error. If you want to add a check for using an incorrect string please create a new test file.

jmmut: I was just testing how it failed, I forgot to restore this file. I will restore it now.

src/vcf/vcf_v41.ragel
jmmut: Not sure I understand, this is done like this because both  `ParsePolicy::handle_body_line` and `OptionalPolicy::optional_check_body_entry` raise a `BodySectionError`, but the first has to be handled as an error and the second is handled as a warning.

inc/vcf/error.hpp
jmmut: and runtime_error, I guess.

src/vcf/store_parse_policy.cpp
jmmut: What about the quality just after this? should we raise a QualityBodyError instead of setting quality to 0?

src/vcf/meta_entry.cpp
jmmut: fixed in next commit.

src/vcf/validate_optional_policy.cpp
jmmut: fixed in next commit

src/vcf/vcf_v41.ragel
cyenyxe: I mean something like this: https://gist.github.com/cyenyxe/e83a47796b369cbd7cfa4854f6c4a0b2

The flow looks a bit more obvious in my opinion.

src/vcf/store_parse_policy.cpp
cyenyxe: Yes, I think we should.

inc/vcf/error.hpp
cyenyxe: The are several others throughout the code. I can point them all if you want me to, but I thought it would make checking the review comments in general a nightmare :)

src/vcf/vcf_v41.ragel
jmmut: Unfortunately, that won't work, that will test the warnings even if there's an error. I'm not changing this, as the current code is the only way I can say that it fully works.

src/vcf/store_parse_policy.cpp
jmmut: oh, tests fail because a '.' is a valid quality. I'll add some code to test it. if quality is '.', are we happy with putting a 0?

cyenyxe: Ugh, I forgot that special case. Yes, please set '.' to 0.

jmmut:
interesting read: https://en.wikibooks.org/wiki/More_C%2B%2B_Idioms/Polymorphic_Exception
it seems that `throw` uses the static type of the object thrown, because it copies it (as the variable will be destroyed with its stack). 

We will have to be careful if we rethrow a `Error`. We may implement that `raise()` interface to be sure.
Here there is a gist showing the issue https://gist.github.com/jmmut/844d3f58aada367e97b23803dbfa99f1

PR #24, opened 2016-05-31
Adding SQLite3 ReportWriter

jmmut:
First version that allows to write reports other than in stdout.
things to note:
- Updated CMakeLists.txt. `-DBUILD_STATIC=0` affects only to boost. SQLite will always be linked statically. 
-  `Error`s are now thrown as pointers, not as references to allow easy copy (exiting the `catch (Error &e)` block was destroying the reference, and no idea how to copy a reference without losing its dynamic type).
- There are `SqliteReportWriter` and `StdoutReportWriter`, both inheriting `ReportWriter`. You can choose any number from 1 to `2` output types from CLI with `-o` option: `-o sqlite,stdout`. I'm not sure about a 3rd output type "silent" that outputs no errors.
- `ParsingState` holds the current line (recored) and now holds as well errors and warnings of the current record.
- To be able to iterate the errors from `main.cpp`, getters had to be added to `Parser` and `ParserImpl`: `errors()` and `warnings()`
- There are a couple of example usage/benchmarks for SQLite.
- There is a minimal test that reads the sqlite ouput


CMakeLists.txt
```diff
@@ -21,9 +22,15 @@ endmacro (declare_mod)
...
+  set(Boost_USE_STATIC_LIBS ON) # only find static libs
+endif (BUILD_STATIC)
 find_package (Boost COMPONENTS filesystem program_options regex system REQUIRED )
 include_directories (${Boost_INCLUDE_DIR} )
 
+add_library(sqlite3 lib/sqlite/sqlite3.c)
+find_package (Threads REQUIRED)
+
 # Static build extra flags
 if (BUILD_STATIC)
```
cyenyxe: The file could seem more structured by moving the "Static build extra flags" before "Dependency libraries", so both `BUILD_STATIC` blocks were merged into one.

inc/vcf/parsing_state.hpp
```diff
@@ -40,6 +41,8 @@ namespace ebi
         
         std::shared_ptr<Source> source;
         std::shared_ptr<std::vector<Record>> records;
+        std::shared_ptr<std::vector<std::shared_ptr<Error>>> errors;
+        std::shared_ptr<std::vector<std::shared_ptr<Error>>> warnings;
```
cyenyxe: Do these need to be shared_ptr? Couldn't they be unique_ptr, since they only belong to the container?

cyenyxe:
The files inside the `sqlitetest` folder shouldn't be in this repository. Since they don't include any headers from the main application, they could go into the [repository for examples](https://github.com/EBIvariation/examples).

src/main.cpp
```diff
@@ -25,59 +25,62 @@
...
     {
         po::options_description description("Usage: vcf-validator [OPTIONS] [< input_file]\nAllowed options");
-        
+
         description.add_options()
             ("help,h", "Display this help")
             ("input,i", po::value<std::string>()->default_value("stdin"), "Path to the input VCF file, or stdin")
             ("level,l", po::value<std::string>()->default_value("warning"), "Validation level (error, warning, stop)")
             ("version,v", po::value<std::string>(), "VCF fileformat version to validate the file against (v4.1, v4.2, v4.3)")
+            ("output,o", po::value<std::string>()->default_value("stdout"), "Comma separated values for type of output (sqlite, stdout, silent)")
```
cyenyxe: Not sure most of the users will understand what "sqlite" means. Maybe we could use "database"?

inc/vcf/parsing_state.hpp
jmmut: which ones? the pointer to vector or the pointer to error? I guess the error one

src/main.cpp
```diff
@@ -100,17 +103,44 @@ namespace
...
+
+        for (auto out : outs) {
+            if (out == "sqlite") {
+                outputs.push_back(std::make_shared<ebi::vcf::SqliteReportWriter>(input + ".errors.db"));
+            } else if (out == "stdout") {
+                outputs.push_back(std::make_shared<ebi::vcf::StdoutReportWriter>());
+            } else if (out == "silent") {
+                ; // do nothing, leaving an empty vector if necessary
+            } else {
+                throw std::invalid_argument{"Please use only valid outputs"};
```
cyenyxe: Use the same message structure as in the rest of errors: "Please choose one of the accepted xxx"

inc/vcf/parsing_state.hpp
cyenyxe: Yes, i mean the pointer to error.

src/vcf/report_writer.cpp
```diff
@@ -0,0 +1,120 @@
+//
+// Created by jmmut on 25/05/16.
+//
```
cyenyxe: License header.

src/vcf/report_writer.cpp
```diff
@@ -0,0 +1,120 @@
...
+#include "vcf/report_writer.hpp"
+
+namespace ebi
+{
+  namespace vcf
+  {
+    SqliteReportWriter::SqliteReportWriter(std::string db_name) : db_name{db_name}, current_transaction_size{0},
+                                                      transaction_size{1000000}, sleep_time{500}
+    {
+        int rc = sqlite3_open(db_name.c_str(), &db);
```
cyenyxe: It should be checked whether the database already exists (probably from a previous run).

src/main.cpp
jmmut: I would go for "database" or the flag "--persistent-report" or similar

CMakeLists.txt
jmmut: done in next commit

src/vcf/report_writer.cpp
```diff
@@ -0,0 +1,120 @@
...
+        if (rc != SQLITE_OK) {
+            sqlite3_close(db);
+            throw std::runtime_error(std::string("Can't open database: ") + sqlite3_errmsg(db));
+        }
+
+        char *zErrMsg = NULL;
+        rc = sqlite3_exec(db, "CREATE TABLE if not exists Errors (line int, message varchar(255));", NULL, 0,
+                          &zErrMsg);
```
cyenyxe: Table and column names in SQL are normally lowercase to avoid case-sensitivity problems.

src/main.cpp
jmmut: Initially it was that way, but the it works now is allowing several types, so not really "choose one".

src/vcf/report_writer.cpp
jmmut: fixed in next commit

src/vcf/report_writer.cpp
```diff
@@ -0,0 +1,120 @@
...
+        char *zErrMsg = NULL;
+        int rc = sqlite3_exec(db, "BEGIN", NULL, 0, &zErrMsg);
+        if (rc != SQLITE_OK) {
+            std::string error_message{std::string("Can't start transaction: ") + zErrMsg};
+            sqlite3_free(zErrMsg);
+            throw std::runtime_error{error_message};
+        }
+    }
+
+    void SqliteReportWriter::finish_transaction()
```
cyenyxe: This method name can be confusing: transactions are either committed or rollbacked, but not "finished". Both possibilities should be supported by the class.

src/vcf/report_writer.cpp
```diff
@@ -0,0 +1,120 @@
...
+            sqlite3_free(zErrMsg);
+            throw std::runtime_error{error_message};
+        }
+    }
+
+    SqliteReportWriter::~SqliteReportWriter()
+    {
+        // in case the amount of inserts is not a multiple of the transaction size, do the leftover inserts in a smaller transaction
+        if (current_transaction_size != 0) {
+            finish_transaction();
```
cyenyxe: It is not a destructor task to commit transactions, but just to free resources. They should have been committed before, and if one was still pending it should be rollbacked.

src/vcf/report_writer.cpp
jmmut: what would be the expected behaviour when a previous (filled) table exists? throw an error? empty the table with a warn?

src/vcf/report_writer.cpp
jmmut: fixed in next commit

test/vcf/report_writer_test.cpp
```diff
@@ -0,0 +1,137 @@
...
+              output.write_warning(*warn);
+          }
+      }
+
+      validator.end();
+
+      return validator.is_valid();
+  }
+
+  TEST_CASE("output errors to sqlite", "[output]")
```
cyenyxe: Before the integration test already implemented, it would be interesting to have unit tests that validate the methods from the writer interface in an isolated way: opening the database, writing an error/warning, etc.

src/main.cpp
jmmut: using "database" for the moment in next commit. will keep an eye on usage to see if the flag fits better

src/vcf/report_writer.cpp
jmmut: done in next commit

src/vcf/report_writer.cpp
jmmut: currently, we don't provide an api of transactions but we do provide a `close()` in next commit. if the connection is not closed, the destructor calls close (just like an ofstream). In case any write or any commit fails, a rollback is performed.

test/vcf/report_writer_test.cpp
```diff
@@ -78,8 +78,73 @@ namespace ebi
...
+              }
+              return 0;
+          }, &count_errors, &zErrMsg);
+          if (rc != SQLITE_OK) {
+              std::string error_message = std::string("Can't read database: ") + zErrMsg;
+              sqlite3_free(zErrMsg);
+              throw std::runtime_error(error_message);
+          }
+          
+          CHECK(count_errors == 2);
```
cyenyxe: Could also check that no warning are inserted.

test/vcf/report_writer_test.cpp
```diff
@@ -78,8 +78,73 @@ namespace ebi
...
+              }
+              return 0;
+          }, &count_warnings, &zErrMsg);
+          if (rc != SQLITE_OK) {
+              std::string error_message = std::string("Can't read database: ") + zErrMsg;
+              sqlite3_free(zErrMsg);
+              throw std::runtime_error(error_message);
+          }
+
+          CHECK(count_warnings == 1);
```
cyenyxe: Could also check that no errors are inserted.

test/vcf/report_writer_test.cpp
jmmut: ok, it's much easier in newer versions, so I'll check it there.

inc/vcf/report_writer.hpp
```diff
@@ -0,0 +1,80 @@
...
+namespace ebi
+{
+  namespace vcf
+  {
+    class ReportWriter
+    {
+      public:
+//        virtual ~ReportWriter() {}  // needed if using raw pointers, instead of references or shared_ptrs
+        virtual void write_error(Error &error) = 0;
+        virtual void write_warning(Error &error) = 0;
```
cyenyxe: const argument maybe?

inc/vcf/report_writer.hpp
```diff
@@ -0,0 +1,80 @@
...
+    };
+
+    class StdoutReportWriter : public ReportWriter
+    {
+      public:
```
cyenyxe: No need for this public section.

inc/vcf/report_writer.hpp
```diff
@@ -0,0 +1,80 @@
...
+        void write(Error &error, std::string table);
+        void start_transaction();
+        void commit_transaction();
+
+        sqlite3* db;
+        std::string db_name;
+        size_t current_transaction_size;
+        const size_t transaction_size;
+        const std::chrono::milliseconds sleep_time;
+        void rollback_transaction();
```
cyenyxe: Would look better moved along the start and commit methods.

src/main.cpp
```diff
@@ -25,59 +25,62 @@
 
 #include "vcf/file_structure.hpp"
 #include "vcf/validator.hpp"
+#include "vcf/report_writer.hpp"
 
+// TODO any reason why this namespace is indented 4 spaces?
```
cyenyxe: IDE formatting xD

src/main.cpp
```diff
@@ -100,17 +103,44 @@ namespace
...
+        size_t initial_size = outs.size();
+
+        // don't write several times to the same output
+        std::sort(outs.begin(), outs.end());
+        std::unique(outs.begin(), outs.end());
+        if (initial_size != outs.size()) {
+            std::cerr << "Warning, duplicated outputs! will write just once to each output specified by -o/--output" << std::endl;
+        }
+
+        std::vector<std::shared_ptr<ebi::vcf::ReportWriter>> outputs;
```
cyenyxe: Couldn't this be unique_ptr?

src/vcf/report_writer.cpp
cyenyxe: In order to prevent overwriting any previous output by mistake, and taking into account the conversation we had some days ago about generating a timestamped file for every run, throwing an exception seems like the way to go.

src/vcf/report_writer.cpp
```diff
@@ -0,0 +1,165 @@
...
+            if (rc != SQLITE_OK) {
+                int remaining_tries = 10;
+                while (rc == SQLITE_BUSY && remaining_tries-- > 0) {
+                    // maybe sqlite is still writing the last transaction
+                    std::cerr << "waiting to finish transactions" << std::endl;
+                    std::this_thread::sleep_for(sleep_time);
+                    rc = sqlite3_close(db);
+                }
+                if (rc != SQLITE_OK) {
+                    throw std::runtime_error{"can not close database " + db_name};
```
cyenyxe: Start with capital letter.

src/vcf/report_writer.cpp
```diff
@@ -0,0 +1,165 @@
...
+            db = nullptr;
+        }
+    }
+
+    SqliteReportWriter::~SqliteReportWriter()
+    {
+        try {
+            close();
+        } catch (std::runtime_error e) {
+            std::cerr << "Error at ~SqliteReportRW: " << e.what() << std::endl;
```
cyenyxe: Maybe we could try to make this error clearer when notified to the user? Something like "An error occurred when finalizing the report generation", using a pun with the meaning of "finalizing".

cyenyxe:
Some constructors use braces and others brackets (see std::string objects nested in SqliteReportWriter class).

src/vcf/report_writer.cpp
jmmut: fixed in next commit

test/vcf/report_writer_test.cpp
jmmut: fixed in newer commit

test/vcf/report_writer_test.cpp
jmmut: same as previous one

inc/vcf/report_writer.hpp
jmmut: oops, why does it work with private interface implementation? I think it should be public, right?

inc/vcf/report_writer.hpp
jmmut: fixed in newer commits

src/main.cpp
jmmut: I would go with coherent indentation, but those changes are a pain for reviewing... I guess it is not the end of the world this way

cyenyxe: Yes, you're right about coherent indentation, I just didn't noticed at first. After all this has been merged a specific commit can be created just for this so it doesn't mess up with the rest of changes.

src/vcf/report_writer.cpp
jmmut: fixed in next commit

src/main.cpp
jmmut: fixed in next commit

src/vcf/report_writer.cpp
jmmut: sigh, I want a compiler error for miscapitalizations... fixed in next commit

jmmut:
parenthesis removed from constructors in next commit

inc/vcf/report_writer.hpp
jmmut: fixed in next commit

inc/vcf/report_writer.hpp
jmmut: I let it public

PR #25, opened 2016-06-07
EVA-341 Reading sqlite reports

jmmut:
improvements over #24:
- added a ReportReader interface.
- SqliteReportRW (renamed from SqliteReportWriter) now implements both ReportReader and ReportWriter.
- added error codes to Error hierachy to instatiate the correct class while reading from DB.
- added more tests for SqliteReporRW.
- The report reader/writer add a virtual destructor, which fixes a bug in the current develop branch, which doesn't flush the writing

### DIFF
--- a/inc/vcf/error.hpp
+++ b/inc/vcf/error.hpp
@@ -24,6 +24,25 @@ namespace ebi
 {
   namespace vcf
   {
+    enum class ErrorCode {
+        error,
+        meta_section,
+        header_section,
+        body_section,
+        fileformat,
+        chromosome_body,
+        position_body,
+        id_body,
+        reference_allele_body,
+        alternate_alleles_body,
+        quality_body,
+        filter_body,
+        info_body,
+        format_body,
+        samples_body,
+    };
+    
+    
     // parent of vcf errors
     class Error : public std::runtime_error
     {
@@ -36,14 +55,9 @@ namespace ebi
                   message{message} {}
 
 
-        size_t get_line() const
-        {
-          return line;
-        }
-        const std::string &get_raw_message() const
-        {
-          return message;
-        }
+        size_t get_line() const { return line; }
+        const std::string &get_raw_message() const { return message; }
+        virtual ErrorCode get_code() const { return ErrorCode::error; }
 
       private:
         size_t line;
@@ -56,6 +70,7 @@ namespace ebi
       public:
         using Error::Error;
         MetaSectionError(size_t line) : MetaSectionError{line, "Error in meta-data section"} { }
+        virtual ErrorCode get_code() const override { return ErrorCode::meta_section; }
     };
 
     class HeaderSectionError : public Error
@@ -63,6 +78,7 @@ namespace ebi
       public:
         using Error::Error;
         HeaderSectionError(size_t line) : HeaderSectionError{line, "Error in header section"} { }
+        virtual ErrorCode get_code() const override { return ErrorCode::header_section; }
     };
 
     class BodySectionError : public Error
@@ -70,6 +86,7 @@ namespace ebi
       public:
         using Error::Error;
         BodySectionError(size_t line) : BodySectionError{line, "Error in body section"} { }
+        virtual ErrorCode get_code() const override { return ErrorCode::body_section; }
     };
 
     // inheritance siblings about detailed errors
@@ -78,6 +95,7 @@ namespace ebi
       public:
         using MetaSectionError::MetaSectionError;
         FileformatError(size_t line) : FileformatError{line, "Error in file format section"} { }
+        virtual ErrorCode get_code() const override { return ErrorCode::fileformat; }
     };
 
     class ChromosomeBodyError : public BodySectionError
@@ -86,6 +104,7 @@ namespace ebi
         using BodySectionError::BodySectionError;
         ChromosomeBodyError(size_t line) : ChromosomeBodyError{line,
             "Chromosome is not a string without colons or whitespaces, optionally wrapped with angle brackets (<>)"} { }
+        virtual ErrorCode get_code() const override { return ErrorCode::chromosome_body; }
     };
 
     class PositionBodyError : public BodySectionError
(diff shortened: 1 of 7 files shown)


inc/vcf/error.hpp
```diff
@@ -36,14 +55,9 @@ namespace ebi
...
-        {
-          return line;
-        }
-        const std::string &get_raw_message() const
-        {
-          return message;
-        }
+        size_t get_line() const { return line; }
+        const std::string &get_raw_message() const { return message; }
+        virtual ErrorCode get_code() const { return ErrorCode::error; }
```
cyenyxe: Could this be made an attribute, same as `line` and `message`, and initialized in the constructor? It would be more consistent with the rest of the class and also prevent virtual calls.

src/vcf/sqlite_report.cpp
```diff
@@ -0,0 +1,310 @@
...
+{
+  namespace vcf
+  {
+    SqliteReportRW::SqliteReportRW(std::string db_name) : db_name{db_name}, current_transaction_size{0},
+                                                          transaction_size{1000000}, sleep_time{500}
+    {
+        int rc = sqlite3_open(db_name.c_str(), &db);
+        if (rc != SQLITE_OK) {
+            sqlite3_close(db);
+            throw std::runtime_error(std::string{"Can't open database: "} + sqlite3_errmsg(db));
```
cyenyxe: Curly braces instead of brackets.

src/vcf/sqlite_report.cpp
```diff
@@ -0,0 +1,310 @@
...
+    }
+    
+    void SqliteReportRW::close()
+    {
+        flush();
+        
+        if (db != nullptr) {
+            int rc = sqlite3_close(db);
+            if (rc != SQLITE_OK) {
+                int remaining_tries = 10;
```
cyenyxe: `unsigned int` to guarantee it will never be less than zero?

src/vcf/sqlite_report.cpp
```diff
@@ -0,0 +1,310 @@
...
+        if (current_transaction_size == 0) {
+            start_transaction();
+        }
+        std::stringstream ss;
+        ss << "INSERT INTO " << table << " ( line , code , message ) VALUES (";
+        ss << error.get_line() << " , "
+        << static_cast<int>(error.get_code()) << " , \""
+        << error.get_raw_message() << "\");";
+
+        rc = sqlite3_exec(db, ss.str().c_str(), NULL, 0, &zErrMsg);
```
cyenyxe: This makes SQLite3 compile a different statement every time an insert is executed, and types are not checked at the application level. https://www.sqlite.org/c3ref/bind_blob.html should be used to bind the parameters, and as a result `sqlite3_prepare_v2` instead of `sqlite3_exec`.

inc/vcf/sqlite_report.hpp
```diff
@@ -0,0 +1,75 @@
...
+        // ReportWriter implementation
+        void write_error(const Error &error) override;
+        void write_warning(const Error &error) override;
+        
+        // ReportReader implementation
+        size_t count_errors() override;
+        size_t count_warnings() override;
+        
+        void for_each_error(std::function<void(std::shared_ptr<Error>)> user_function);
+        void for_each_warning(std::function<void(std::shared_ptr<Error>)> user_function);
```
cyenyxe: The compiler raises a warning `'for_each_xxx' overrides a member function but is not marked 'override'`.

test/vcf/report_writer_test.cpp
```diff
@@ -140,63 +141,141 @@ namespace ebi
...
-              CHECK_FALSE(is_valid(path.string(), *output));
-          }
-          int rc = sqlite3_open(db_name.c_str(), &db);
-          if(rc != SQLITE_OK) {
-              sqlite3_close(db);
-              throw std::runtime_error(std::string("Can't open database: ") + sqlite3_errmsg(db));
-          }
+          std::shared_ptr<ebi::vcf::ReportWriter> output{std::make_shared<ebi::vcf::SqliteReportRW>(db_name)};
+          CHECK_FALSE(is_valid(path.string(), *output));
+      }
```
cyenyxe: Nesting no longer necessary.

test/vcf/report_writer_test.cpp
```diff
@@ -140,63 +141,141 @@ namespace ebi
...
+      SECTION(path.string() + " error count")
+      {
+          size_t count_errors;
+          size_t count_warnings;
+
+          {
+              ebi::vcf::SqliteReportRW errorsDAO{db_name};
+              count_errors = errorsDAO.count_errors();
+              count_warnings = errorsDAO.count_warnings();
           }
```
cyenyxe: Unnecessary nesting, the counters could be declared and assigned at once.

inc/vcf/error.hpp
jmmut: yes, I didn't think about it, it makes the code a bit more elegant. fixed in next commit

src/vcf/sqlite_report.cpp
jmmut: fixed

inc/vcf/error.hpp
jmmut: oops, running tests, it seems that it's not so easy. when using the children 2-argument constructor, we would need to add another constructor like 
```
class FileformatError : public MetaSectionError
{
  public:
    // using MetaSectionError::MetaSectionError; // compilation error: constructor with 2 arguments collides
    FileformatError(size_t line) : FileformatError{line, "Error in file format section"} { }
    FileformatError(size_t line, std::string message) : FileformatError{ErrorCode::fileformat, line, message} { }// compilation error: there's no FileformatError that takes 3 arguments
};
```
and removing the `using Error::Error` as we are redefining the 2-argument constructor, but then the 3-argument is not accesible with the same name, nor even with the parent's name (MetaSectionError in this example), and calling a non-direct parent is not allowed, so we can not even `FileformatError(size_t line, std::string message) : Error{ErrorCode::fileformat, line, message} { }`.

Unless I did something very silly, I would let this as it was before.

src/vcf/sqlite_report.cpp
jmmut: I made it the more intuitive way I could. I am using a postdecrement, in the next condition, so a `remaining_tries = 10` means that it will sleep 10 times. in the last iteration, it is doing:
```
// remaining_tries = 1;
while (... remaining_tries-- > 0){   // 1 > 0 true
    // remaining_tries = 0;
    ... // last wait
}
while (... remaining_tries-- > 0){ // 0 > 0 false
} // exit loop
// remaining_tries = -1; // but it is not used anymore
```
using `unsigned` leads either a less intuitive loop of less intuitive counts. or we can use a `for` loop of a condition `rc == SQLITE_BUSY && i < remaining_tries`

cyenyxe: Oh yeah, I understand what the loop does, maybe I was just being paranoid with type safety.

src/vcf/sqlite_report.cpp
jmmut: fixed, times are now ~95% of the previous times.

test/vcf/report_writer_test.cpp
jmmut: The extra block is needed to flush, SqliteReportRW provides a RAII-style for flushing the writings. It provides `close()` and `flush()` as well, but we are handling the interface ReportWriter here. To remove the block we can add `close()` and `flush()` methods to the ReportWriter interface.

test/vcf/report_writer_test.cpp
jmmut: same as last comment

inc/vcf/sqlite_report.hpp
jmmut: fixed

PR #26, opened 2016-06-10
EVA-337 adding timestamp and outdir, and removing "silent" report

jmmut:


src/main.cpp
```diff
@@ -46,7 +49,8 @@ namespace
             ("input,i", po::value<std::string>()->default_value("stdin"), "Path to the input VCF file, or stdin")
             ("level,l", po::value<std::string>()->default_value("warning"), "Validation level (error, warning, stop)")
             ("version,v", po::value<std::string>(), "VCF fileformat version to validate the file against (v4.1, v4.2, v4.3)")
-            ("output,o", po::value<std::string>()->default_value("stdout"), "Comma separated values for type of output (database, stdout, silent)")
+            ("report,r", po::value<std::string>()->default_value("stdout"), "Comma separated values for types of reports (database, stdout)")
+            ("outdir,o", po::value<std::string>()->default_value(""), "Directory for the output (same as input file)")
```
cyenyxe: What does this message mean? Is "same as input file" the default value? The rest of clarifications between brackets are not the defaults. Also, when I run the program it uses the current directory, not the one where the input file is contained.

src/main.cpp
```diff
@@ -108,6 +112,20 @@ namespace
...
+        
+        boost::filesystem::path file_boost_path{file_path};
+        boost::filesystem::path outdir_boost_path{outdir};
+        
+        outdir_boost_path /= file_boost_path.filename();
+        
+        return outdir_boost_path.string();
+    }
+
     std::vector<std::unique_ptr<ebi::vcf::ReportWriter>> get_outputs(std::string const &output_str, std::string const &input) {
```
cyenyxe: The warning message this function displays should refer to reports and the "-r/--report" argument now.

src/main.cpp
```diff
@@ -124,18 +142,22 @@ namespace
 
         for (auto out : outs) {
             if (out == "database") {
-                std::string db_filename = input + ".errors.db";
+//                time_t now = std::chrono::system_clock::to_time_t(std::chrono::system_clock::now());
+//                std::stringstream ss;
+//                std::cout << input << ".errors." << std::put_time(std::localtime(&now), "%F") << ".db";
+//                std::string db_filename = ss.str();
```
cyenyxe: Remove commented block.

src/main.cpp
jmmut: Do I add an explanation in the message? or leave it as just "Directory for the output" and develop the explanation in the readme?

jmmut: BTW, we should update the readme anyways, it doesn't say that the version is mandatory

cyenyxe: If both cases are explained it will get quite long, so probably better to leave it for the readme file.

src/main.cpp
jmmut: fixed in next commit

src/main.cpp
jmmut: fixed in next commit

PR #27, opened 2016-06-15
cmake: added explicit list of sources

jmmut:
This avoids some compilation errors when adding or renaming
files, and sharing the code between several developers. This
way, there is no need for "extra, just in case" clean rebuilds.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -11,15 +11,35 @@ set (CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/bin)
 include_directories (inc)
 include_directories (lib)
 
-macro (declare_mod MODNAME)
-  file (GLOB MOD_${MODNAME}_INC "inc/${MODNAME}/*")
-  file (GLOB MOD_${MODNAME}_SRC "src/${MODNAME}/*")
-  set (MOD_${MODNAME}_ALL ${MOD_${MODNAME}_INC} ${MOD_${MODNAME}_SRC})
-  add_library (mod_${MODNAME} ${MOD_${MODNAME}_ALL})
-endmacro (declare_mod)
-
 # Application modules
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -11,15 +11,48 @@ set (CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/bin)
...
-macro (declare_mod MODNAME)
-  file (GLOB MOD_${MODNAME}_INC "inc/${MODNAME}/*")
-  file (GLOB MOD_${MODNAME}_SRC "src/${MODNAME}/*")
-  set (MOD_${MODNAME}_ALL ${MOD_${MODNAME}_INC} ${MOD_${MODNAME}_SRC})
-  add_library (mod_${MODNAME} ${MOD_${MODNAME}_ALL})
-endmacro (declare_mod)
-
 # Application modules
-declare_mod (vcf)
+set (MOD_VCF_SOURCES
```
cyenyxe: Is it possible to list these by folder at least?

jmmut: I would say not, the problem would remain when we add files.

jmmut: > The problem with our current approach (do a module of everything under src/vcf) is that when files are added or removed across commits (specially when coworkers pull changes) without changing CMakeLists.txt, the generator (e.g. make) does not rerun cmake (as the CMakeLists.txt's timestamp is not newer), so the old generator's targets remain, without adding the new targets that in fact there are in the folders.

cyenyxe: What I mean is grouping the explicit list, first with the files inside `inc` and then `src`. Right now they are all mixed.

jmmut: Or are you talking about sorting the lines? I interleaved them by class.

jmmut: fixed

PR #28, opened 2016-06-15
EVA-373: scaffolding debugulator_main

jmmut:
It is not adding functionality, but only creates another main and sets the place for its main loop. This PR can also serve to decide some naming and parameter conventions before developing more.

### DIFF
--- a/src/debugulator_main.cpp
+++ b/src/debugulator_main.cpp
@@ -1,0 +1,290 @@
+/**
+ * Copyright 2014-2015 EMBL - European Bioinformatics Institute
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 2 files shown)


src/debugulator_main.cpp
```diff
@@ -0,0 +1,290 @@
...
+
+  enum class ValidationLevel
+  {
+      error, warning, stop
+  };
+
+
+  po::options_description build_command_line_options()
+  {
+      po::options_description description("Usage: vcf-validator [OPTIONS] [< input_file]\nAllowed options");
```
cyenyxe: Change name of the binary to vcf-debugulator.

src/debugulator_main.cpp
```diff
@@ -0,0 +1,290 @@
+/**
+ * Copyright 2014-2015 EMBL - European Bioinformatics Institute
```
cyenyxe: Copyright is just 2016.

PR #29, opened 2016-06-16
EVA-339 Simple normalization

jmmut:
Preparing variant normalization to be able to detect duplicates.

Given a Record, the function `normalize` returns a vector of RecordCore, which is just the main fields in a Record.

The normalization performs two actions:
- split the multiallelic variants into several individual ones. They are NOT sorted by position, but by allele as listed in the original variant.
- the repeated bases (which serve as leading or trailing context) in both reference and alternate alleles are removed.

Questions:
- Do we use a NormalizationError if a precondition is violated? currently using just BodySectionError.
- This was the first prototype of `normalize`, so I was looking at how much can be reused across the cases of insertions, deletions and polymorphisms. I think the code may be reused a bit more, but maybe making it less understandable. We can decide how much to reuse and split it into functions.


jmmut:
This is what I'm talking about reusing code: https://github.com/jmmut/vcf-validator/commit/589370653db34b5ba1d9a9f50371f9a3487626ff

inc/util/string_utils.hpp
```diff
@@ -46,6 +46,21 @@ namespace ebi
 
         output.swap(ret);
     }
+
+    /**
+     * Temporal implementation for mismatch with 2 starts and 2 ends. It is not in STL for c++11, but it will in c++14.
```
cyenyxe: From what I see in [Clang](http://clang.llvm.org/cxx_status.html#cxx14) and [GCC](https://gcc.gnu.org/projects/cxx-status.html#cxx14)'s standard support documentation, all the compilers we test our code against should be able to handle C++14. Maybe we could consider an upgrade in next release?

inc/vcf/record_cache.hpp
```diff
@@ -0,0 +1,59 @@
...
+                    std::stringstream ss;
+                    ss << "Duplicated record detected: first occurrence: " << *unique.first 
+                    << ", second occurrence: " << record_core;
+                    throw new BodySectionError{record_core.line, ss.str()};
+                }
+            }
+        }
+        
+      private:
+        size_t capacity;
```
cyenyxe: Compiler warning (Clang 3.5): `private field 'capacity' is not used`

inc/vcf/record_cache.hpp
```diff
@@ -0,0 +1,59 @@
...
+            for (RecordCore &record_core: record_cores) {
+                std::pair<std::set<RecordCore>::iterator, bool> unique = cache.insert(record_core);
+                if(not unique.second) {
+                    std::stringstream ss;
+                    ss << "Duplicated record detected: first occurrence: " << *unique.first 
+                    << ", second occurrence: " << record_core;
+                    throw new BodySectionError{record_core.line, ss.str()};
+                }
+            }
+        }
```
cyenyxe: Compiler warning (Clang 3.5): `control may reach end of non-void function`. The function should return a ReturnCode object (or a list of them?) but there's no return statement.

inc/util/stream_utils.hpp
```diff
@@ -40,6 +44,43 @@ namespace ebi
...
+    template <typename T>
+    std::ostream &print_container(std::ostream &os, const T &container) {
+        size_t size = container.size();
+        os << "[";
+        if (size > 0) {
+            auto it = container.begin();
+            os << *it;
+            it++;
+            for (; it != container.end(); ++it) {
+                os << ", " << *it;
```
cyenyxe: Compiler error (Clang 3.5): `call to function 'operator<<' that is neither visible in the template definition nor found by argument-dependent lookup`

inc/vcf/normalizer.hpp
```diff
@@ -0,0 +1,90 @@
...
+                   const std::string &reference_allele, const std::string &alternate_alleles) 
+            : line(line), chromosome(chromosome), position(position), 
+              reference_allele(reference_allele), alternate_alleles(alternate_alleles)
+        { }
+        
+        /** A record "a" is less than another "b" iff:
+         * - the chromosome string from "a" is less than the chromosome string from "b", or if they equal:
+         * - the position from "a" is less than the position from "b", or if they equal:
+         * - the reference_allele string from "a" is less than the reference_allele string from "b", or if they equal:
+         * - the alternate_alleles string from "a" is less than the alternate_alleles string from "b".
```
cyenyxe: It would be useful to differentiate when lexicographical or numerical sorting is used in each of these points.

inc/vcf/parsing_state.hpp
```diff
@@ -43,7 +45,8 @@ namespace ebi
         std::shared_ptr<std::vector<Record>> records;
         std::shared_ptr<std::vector<std::unique_ptr<Error>>> errors;
         std::shared_ptr<std::vector<std::unique_ptr<Error>>> warnings;
-     
+        RecordCache duplicates;
```
cyenyxe: Is this variable used? I only see it commented in `parsing_state.cpp`.

inc/vcf/record_cache.hpp
```diff
@@ -0,0 +1,59 @@
...
+  {
+    
+    class RecordCache
+    {
+      public:
+
+        RecordCache() { }
+        RecordCache(size_t capacity) : capacity(capacity) { }
+
+        RecordCore check(const Record &record)
```
cyenyxe: A name like `check_duplicates` would make the purpose of this method clearer.

src/vcf/normalizer.cpp
```diff
@@ -0,0 +1,189 @@
...
+            corrected_position = position + (lead_mismatch_indices.first - reference.begin());
+
+            records.emplace_back(record.line, record.chromosome, corrected_position,
+                                 corrected_reference, corrected_alternate);
+        }
+
+        return std::move(records);
+    }
+    
+    std::vector<RecordCore> normalize_pad_at_left(const Record &record/* , ParsingState?*/)
```
cyenyxe: The English term seems to be "left alignment" and not "left padding". Padding would probably mean the opposite :)

inc/util/string_utils.hpp
jmmut: but is the support stable or experimental? I would try to stick with at least the default deployed version with the distros.

Although, now I remember, gcc 4.8 is not c++11-library compliant, so... it invalidates my point. sigh

inc/vcf/record_cache.hpp
jmmut: fixed

inc/vcf/record_cache.hpp
jmmut: not necessary for this version. will put `void` and later we can decide about reporting several errors at the same time

inc/util/stream_utils.hpp
jmmut: really? it does compile in my gcc and in Travis'... will install clang as well...
Is this a bug in gcc?

inc/util/string_utils.hpp
cyenyxe: For Clang, C++14 is supported from version 3.4, released in Jan 2014 (current 3.8).
For GCC, version 4.9 supports most of the features and 5.x the rest. Support for C++17 is experimental.

inc/util/stream_utils.hpp
cyenyxe: gcc tends to be more "relaxed" with spec validations, thus the reason why Clang was included in Travis. [Their APT repo](http://llvm.org/apt/) seems to be up again!

src/vcf/normalizer.cpp
jmmut: actually, I meant that the context is at the left, this is the VCF spec compliant (not the common non-compliant (!) left-alignment), so I guess it would be clearer as normalize_right_alignment

inc/vcf/record_cache.hpp
jmmut: fixed in next commit

src/vcf/normalizer.cpp
cyenyxe: Oops yes, you're right :sweat_smile:

inc/vcf/parsing_state.hpp
jmmut: Mmmh, no, it's not used. It helped me to design the `normalize` interface, I will remove it.

inc/vcf/normalizer.hpp
jmmut: fixed in next commit

src/vcf/normalizer.cpp
```diff
@@ -0,0 +1,189 @@
...
+            // using the previous iterators, assign only [first_mismatch, last_mismatch+1)
+            corrected_reference.assign(lead_mismatch_indices.first, trail_match_ref);
+            corrected_alternate.assign(lead_mismatch_indices.second, trail_match_alt);
+            corrected_position = position + (lead_mismatch_indices.first - reference.begin());
+
+            records.emplace_back(record.line, record.chromosome, corrected_position,
+                                 corrected_reference, corrected_alternate);
+        }
+
+        return std::move(records);
```
cyenyxe: Looks like this is not necessary/recommended because Clang is raising a warning: `moving a local object in a return statement prevents copy elision`. More info in http://vmpstr.blogspot.co.uk/2015/12/redundant-stdmove.html

PR #32, opened 2016-07-06
EVA-340 Detect and report duplicate variants

jmmut:
To do this we normalize and store every Record in a std::multiset<RecordCore>, where RecordCore is a subset of a Record, which only has the line, chromosome, position, reference and alternate.

By default, the capacity of the multiset is set to 1000 RecordCores, as it should enough if the VCF is sorted, and we will report as an error if it's unsorted. The capacity can be configured to unlimited, but not from the CLI.

Some tests added for the RecordCache, and some invalid test VCF added to test the "unsorted" and "duplicated" reports.


inc/vcf/record_cache.hpp
```diff
@@ -0,0 +1,127 @@
...
+                std::pair<std::multiset<RecordCore>::iterator, std::multiset<RecordCore>::iterator> range =
+                        cache.equal_range(record_core);
+
+                if (range.first == range.second) {
+                    // no matches found
+                } else {
+                    // one or more matches found
+                    std::stringstream ss;
+                    ss << "Duplicated record detected: first occurrence: " << *range.first
+                    << ", second occurrence: " << record_core;
```
cyenyxe: The way the fields are displayed is not very clear for an end-user: the line number and the variant fields should be split instead of displayed all together between brackets. This message could be simplified to something like:

> "Duplicated variant " << record_core << " found in lines " << *range.first.line << " and " << record_core.line

inc/vcf/record_cache.hpp
```diff
@@ -0,0 +1,127 @@
...
+     */
+    class RecordCache
+    {
+      public:
+        RecordCache() : RecordCache{1000} { }
+
+        /**
+         * @param capacity: maximum amount of RecordCores that this instance can hold at any time.
+         * Default is 1000.
+         * A value of 0 is valid but is useless.
```
cyenyxe: If a zero value is useless as it is, maybe it could be moved into the other branch (zero or less disables the limit).

inc/vcf/record_cache.hpp
```diff
@@ -0,0 +1,127 @@
...
+         * A negative value disables the limit, thus storing every RecordCore received, use with caution.
+         */
+        RecordCache(long capacity)
```
cyenyxe: Better to enforce the argument type than casting inside the method.

inc/vcf/validator.hpp
```diff
@@ -96,6 +97,11 @@ namespace ebi
        
       protected:
         virtual void parse_buffer(char const * p, char const * pe, char const * eof) = 0;
+
+        /**
+         * Previously seen records
+         */
+        RecordCache duplicates;
```
cyenyxe: If the cache doesn't store duplicates, wouldn't it be better to rename the variable to something like "previous_records"?

test/vcf/test_utils.hpp
```diff
@@ -0,0 +1,65 @@
...
+      size_t normalized_pos;
+      std::string normalized_reference;
+      std::vector <std::string> normalized_alternate;
+  };
+  struct TestRecord
+  {
+      size_t normalized_pos;
+      std::string normalized_reference;
+      std::string normalized_alternate;
+  };
```
cyenyxe: Is TestRecord used at all now?

src/vcf/store_parse_policy.cpp
```diff
@@ -166,6 +166,8 @@ namespace ebi
             throw new PositionBodyError{state.n_lines};
         }
 
+        check_sorted(state, position);
```
cyenyxe: Should errors due to unsorted records (chrom+pos related) be reported before semantic chrom errors? Maybe both the sorting and duplicates checks could go into ParserImpl. I think we already discussed moving the duplicates at least?

jmmut: I'm not sure I understand. What do you mean with semantic chrom errors? where is that? As I see it, there's only syntax errors (ragel regex) and semantic warnings (meta description) for chrom. the "unsorted error" is treated like a semantic error, and it's done where all the semantic errors are checked, store_parse_policy.

The duplicates check is (already) in the ragel code because of the corner case of returning several errors from the same line. (It is in the ragel part because it can not go into their father ParserImpl because it doesn't know anything about policies, and the children ParserImpl_v4* do. BTW, now I see that the code could be extracted into a method in the ParserImpl_v4* headers 3 times, instead of copied in the ragels 3 times, but it's not a big improvement).

inc/vcf/validator.hpp
jmmut: fixed in next commit

inc/vcf/record_cache.hpp
jmmut: fixed in next commit

inc/vcf/record_cache.hpp
jmmut: fixed in next commit

inc/vcf/record_cache.hpp
jmmut: fixed in next commit

PR #33, opened 2016-07-12
EVA-342 Debugulator: remove duplicated variants

jmmut:
Questions:
- I moved the fix_vcf_file to its own file (unlike the is_valid_vcf_file) to be able to reuse from the test. Is this ok?
- Do we want to log every fix? It's likely that some fixes will be a bit risky, and we might want to `warn` about them. 
- Almost all of the fixes in Fixer are unimplemented, and throw exceptions. Do we want to catch them (in the fix_vcf_file loop) and ignore them (or just log them) for the release? otherwise it's going to be frustrating, as there's no easy way to remove the unfixable errors from the report DB.
- As a detail in the fix_vcf_file loop, a different instance of Fixer is created in every iteration because the `vector<char> line` needs to be stored in the Fixer class. This can be solved in different ways: 1: one instance per iteration and store a `vector<char>&`. 2: one instance for all the iterations and store a `vector<char>*`. 3: I don't think it's a good idea to use `managed_ptr`s pointing to local variables, but in the loop we could use a `managed_ptr<vector<char>>` instead of a local variable.


inc/vcf/error.hpp
```diff
@@ -127,69 +173,87 @@ namespace ebi
...
         using BodySectionError::BodySectionError;
         SamplesBodyError(size_t line) : SamplesBodyError{line, "Error in samples columns, in body section"} { }
         virtual ErrorCode get_code() const override { return ErrorCode::samples_body; }
+        virtual void apply_visitor(ErrorVisitor &visitor) { visitor.visit(*this); }
     };
     class NormalizationError : public BodySectionError
     {
       public:
         using BodySectionError::BodySectionError;
         NormalizationError(size_t line) : NormalizationError{line, "Normalization could not be performed"} { }
```
cyenyxe: Clarify it is the allele normalization, users don't necessarily know.

inc/vcf/error.hpp
```diff
@@ -127,69 +173,87 @@ namespace ebi
...
         virtual ErrorCode get_code() const override { return ErrorCode::normalization; }
+        virtual void apply_visitor(ErrorVisitor &visitor) { visitor.visit(*this); }
+    };
+    class DuplicationError : public BodySectionError
+    {
+      public:
+        using BodySectionError::BodySectionError;
+        DuplicationError(size_t line) : DuplicationError{line, "A duplicated record was found"} { }
```
cyenyxe: record -> variant

src/debugulator_main.cpp
```diff
@@ -14,43 +14,37 @@
...
-      ;
+                         ("help,h", "Display this help")
+                         ("input,i", po::value<std::string>()->default_value("stdin"),
+                          "Path to the input VCF file, or stdin")
+                         ("errors,e", po::value<std::string>(),
+                          "Path to the errors DB from the input VCF file")
+                         ("level,l", po::value<std::string>()->default_value("warning"),
+                          "Validation level (error, warning, stop)")
+                         ("output,o", po::value<std::string>()->default_value("stdout"),
+                          "Write to a file or stdout");
```
cyenyxe: I think indentation got a bit crazy here :)

src/vcf/debugulator.cpp
```diff
@@ -0,0 +1,67 @@
...
+                        std::ostream &output)
+      {
+          std::vector<char> line;
+          line.reserve(default_line_buffer_size);
+
+          size_t current_line = 0;  // the first line is the number 1, ParsingState takes this convention too
+
+          size_t errors = errorDAO.count_errors();
+          if (errors == 0) {
+              throw std::runtime_error{"The errors DB was empty, there are no errors to fix the input"};
```
cyenyxe: I don't think this should raise an error, an empty DB will be the output for any correct VCF file so it's not an erroneous situation. This message could be simply displayed.

inc/vcf/fixer.hpp
```diff
@@ -0,0 +1,117 @@
...
+        }
+        virtual void visit(NormalizationError &error) override
+        {
+            throw std::runtime_error{"visit(NormalizationError &error) unimplemented"};
+        }
+        virtual void visit(DuplicationError &error) override
+        {
+            // TODO better log system, if any
+            std::cerr << "## fixing duplicate: removing line " << line_number << ": "
+            << std::string{line.begin(), line.end()} << std::endl;
```
cyenyxe: This is not an error message, should go to std::cout.

src/debugulator_main.cpp
```diff
@@ -14,43 +14,37 @@
...
-              ("input,i", po::value<std::string>()->default_value("stdin"), "Path to the input VCF file, or stdin")
-              ("errors,e", po::value<std::string>(), "Path to the errors DB from the input VCF file")
-              ("level,l", po::value<std::string>()->default_value("warning"), "Validation level (error, warning, stop)")
-              ("output,o", po::value<std::string>()->default_value("stdout"), "write to a file or stdout")
-      ;
+                         ("help,h", "Display this help")
+                         ("input,i", po::value<std::string>()->default_value("stdin"),
+                          "Path to the input VCF file, or stdin")
+                         ("errors,e", po::value<std::string>(),
+                          "Path to the errors DB from the input VCF file")
```
cyenyxe: DB -> report

src/debugulator_main.cpp
```diff
@@ -69,6 +63,12 @@ namespace
           return 1;
       }
 
+      if (!vm.count("errors")) {
+          std::cout << desc << std::endl;
+          std::cout << "Please specify the path to the errors DB (--errors)" << std::endl;
```
cyenyxe: DB -> report

src/vcf/fixer.cpp
```diff
@@ -0,0 +1,26 @@
...
+#include "vcf/fixer.hpp"
+
+namespace ebi
+{
+  namespace vcf
+  {
+
+  }
+}
+
```
cyenyxe: Is this file necessary? Maybe it had something else inside before?

test/vcf/debugulator_test.cpp
```diff
@@ -0,0 +1,130 @@
...
+              }
+          }
+      }
+
+      validator.end();
+
+      return validator.is_valid();
+  }
+
+  TEST_CASE("Fixing a VCF with duplicates", "[debugulator]")
```
cyenyxe: This test should be more modular, for instance, with a unit test just for the fixing operation, removing the step that generates the report.

cyenyxe:
It would be good to log every fix that's being done,. Logging all the errors that couldn't be fixed might be a bit redundant, so I would suggest ignoring the thrown exceptions and just count the number of them. Then at the end of the execution we could show a message like "10 errors couldn't be automatically fixed".

src/vcf/debugulator.cpp
```diff
@@ -0,0 +1,67 @@
...
+                  current_line++;
+                  if (current_line == line_index) {
+                      break;
+                  }
+                  for (auto c : line) {
+                      output << c;
+                  }
+              }
+
+              ebi::vcf::Fixer{line_index, line, output}.fix(*error);
```
cyenyxe: Maybe the Fixer could be initialized just with the "output" argument, whereas "line_index" and "line" would be arguments of the fix method? They would then be assigned to the Fixer object attributes, and the constructor call could be extracted from the loop.

cyenyxe:
I'm a bit conflicted about the output writing. It seems convenient to write the output VCF into stdout, but then it forces us to write any other message into stderr...

src/vcf/debugulator.cpp
```diff
@@ -0,0 +1,67 @@
...
+                      // file finished, return ?
+                      return;
+                  }
+                  current_line++;
+                  if (current_line == line_index) {
+                      break;
+                  }
+                  for (auto c : line) {
+                      output << c;
+                  }
```
cyenyxe: I'm a bit conflicted about the output writing. Writing the output VCF to stdout definitely seems the most convenient way, but then it forces us to write any other message into stderr...

inc/vcf/error.hpp

inc/vcf/error.hpp

src/debugulator_main.cpp
jmmut: yes, the alignment doesn't seem to be a good policy here. fixed in next commit.

src/vcf/debugulator.cpp
jmmut: this is the last point in the PR description. In order to not copy the whole line:
> ...  2: one instance for all the iterations and store a vector<char>*. 3: I don't think it's a good idea to use managed_ptrs pointing to local variables, but in the loop we could use a managed_ptr<vector<char>> instead of a local variable.

src/vcf/debugulator.cpp
jmmut: fixed in next commit

src/debugulator_main.cpp
jmmut: fixed in next commit

src/debugulator_main.cpp
jmmut: fixed in next commit

src/vcf/fixer.cpp
jmmut: mmh, yes, not needed for the moment. fixed in next commit.

test/vcf/debugulator_test.cpp
jmmut: Do I just add some unit tests, or remove the current as well? I think it's valuable to have the whole process tested.

src/vcf/debugulator.cpp
jmmut: It's worse than that: in debugulator_main, we allow to write the fixed vcf to a file instead of stdout, and in that case, the logs are still written into stderr, and nothing into stdout.

Even with that, I think it's worse to merge vcf lines and messages in stdout. The way it is currently is how I thought it would be better.

jmmut:
log only the fixes done and counting the unfixable errors: fixed in next commit

src/vcf/debugulator.cpp
jmmut: I'm going with the option 2. It's a bit uglier, but has a smaller scope (which makes possible errors to be more location-constrained), and doesn't complicate users of Fixer, which shouldn't care about the internal storage of the line.

```
        void fix(size_t line_number, std::vector<char> &line, Error &error)
        {
            this->line_number = line_number;
            this->line = &line; // this will be valid until this function returns
            error.apply_visitor(*this);
            this->line = nullptr;
        }
private:
        std::vector<char> *line;
```

Otherwise, the users of Fixer need to handle a `std::shared_ptr<std::vector<char>> line`, making more difficult the creation (likely with `make_shared`) and change its uses from `line` to `*line` in every place.

cyenyxe: What about passing it by value? Since it's a vector, I think it would be moved around implicitly.

test/vcf/debugulator_test.cpp
cyenyxe: System tests could be extracted to a different file than unit tests. That would remove redundancy, because there is a whole-process test for the validator side too, and it feels a bit redundant now.

src/vcf/debugulator.cpp
jmmut: I'm not sure about that. Outside in the `fix_vcf_file` loop we use the same vector several times to fill from `readline` (which does a `.clear()`) and passing it to Fixer. I think a move would only be possible if that object was not used anymore. And if it was moved, I don't easily see the consequences of that posterior `.clear()`

PR #34, opened 2016-07-18
INFO AA field now accepts values according to spec - EVA-377

cyenyxe:
The INFO AA field was described in the parser as in 1000 Genomes: http://ftp.1000genomes.ebi.ac.uk/vol1/ftp/phase1/analysis_results/supporting/ancestral_alignments/human_ancestor_GRCh37_e59.README

This is more restrictive than what the specification states officially, so the parser now allows any character except those clearly restricted in it. Versions 4.1/4.2 won't accept white-spaces but 4.3 does.

A non-applied rule for structural variants has been also removed for now.


src/vcf/vcf_v43.ragel
```diff
@@ -721,8 +721,9 @@
                         ( "AA" | "AC" | "AD" | "ADF" | "ADR" | "AF" | "AN" | "BQ" | "CIGAR" | "DB" | "DP" | "END" | "H2" | "H3" | "MQ" | "MQ0" | "NS" | "SB" | "SOMATIC" | "VALIDATED" | "1000G" ) ;
     info_value      = (print - ";")+ ;
     info_value_list = info_value ("," info_value)* ;
+    info_aa         = (print - ("," | ";" | "="))+ ;
     info_entry      = (
```
jmmut: I think we should try to maintain all 3 version the most similar possible. That will help to keep the differences well highlighted.

 I would introduce here `info_entry_snv` as well. Between versions 4.1 and 4.2 there's no difference here. 4.3 allows spaces (and some punctuation things) in `info_value`, `info_key` and `info_aa`, but otherwise than that I don't see more differences.

PR #38, opened 2016-07-22
EVA-401 Adding ODB for mapping Error classes

jmmut:
This PR adds a basic usage of odb for storing Error classes. 

The class OdbReportRW sustitutes our SqliteReporRW, and is already equally featured (check report_writer_test.cpp, "unit tests: odb").

This will allow us to (more easily) add and persist different members in each Error child class, depending on how we want to fix each error.

Although, it is automated in the cmake, the odb compilation is similar to this: `odb --include-prefix vcf --std c++11 -d sqlite --generate-query --generate-schema inc/vcf/error.hpp` and then moving the error-odb.cpp to src/vcf . To compile the odb classes you need the packages `odb`, `libodb` and `libodb-sqlite`.


README.md
```diff
@@ -11,9 +11,17 @@ Please read the wiki for more details about checks already implemented.
...
 ## Dependencies
 
+### Boost
+
 The dependencies are the Boost library core, and its submodules: Boost.filesystem, Boost.program_options, Boost.regex and Boost.system.
 If you are using Ubuntu, the required packages names will be `libboost-dev`, `libboost-filesystem-dev`, `libboost-program-options-dev` and `libboost-regex-dev`.
 
+### ODB
+
+You will need to download the odb compiler, the odb common runtime library, and the sqlite database runtime library from this page: http://codesynthesis.com/products/odb/download.xhtml
```
cyenyxe: Hyperlink missing, it is just plain text now.

README.md
```diff
@@ -11,9 +11,17 @@ Please read the wiki for more details about checks already implemented.
...
+
+To install the odb compiler, the easiest way is to download the `.deb` or `.rpm` version, in order to be installed automatically with `dpkg`. Both the odb runtime and sqlite libraries can be installed with the usual `./configure && make && sudo make install`.
```
cyenyxe: We should be more specific about what refers to the package and what to the manual installation, otherwise I see non-technical users trying to run `configure` on a .deb... Maybe something like:

"If not using these type of packages, both the odb runtime and sqlite libraries can be manually build and installed"

inc/vcf/error.hpp
```diff
@@ -1,4 +1,77 @@
 /**
+ ODB License Exception for vcf-validator (Validator for the Variant Call
```
cyenyxe: This license exception text is huge! It could be moved to the end of the LICENSE file or placed in its own file, but better not to mix it with the code.

inc/vcf/error.hpp
```diff
@@ -113,144 +193,191 @@ namespace ebi
...
+        #pragma db column("COUNT(" + Error::id_ + ")")
+        std::size_t count;
     };
 
     // inheritance siblings depending on file location
+    #pragma db object
     class MetaSectionError : public Error
     {
       public:
+        friend class odb::access;
```
cyenyxe: Apart from `Error`, this is declared only in this class.

README.md
jmmut: only the compiler has a `.deb` option to download, but I agree with anything that improves clarity.

src/main.cpp
```diff
@@ -50,7 +51,7 @@ namespace
             ("input,i", po::value<std::string>()->default_value("stdin"), "Path to the input VCF file, or stdin")
             ("level,l", po::value<std::string>()->default_value("warning"), "Validation level (error, warning, stop)")
             ("version,v", po::value<std::string>(), "VCF fileformat version to validate the file against (v4.1, v4.2, v4.3)")
-            ("report,r", po::value<std::string>()->default_value("stdout"), "Comma separated values for types of reports (database, stdout)")
+            ("report,r", po::value<std::string>()->default_value("stdout"), "Comma separated values for types of reports (database, stdout, orm)")
```
cyenyxe: How compatible are the databases generated by ODB and plain SQLite? People in general won't understand what an "orm" is, so at some point the "database" option should be just replaced.

src/vcf/odb_report.cpp
```diff
@@ -0,0 +1,150 @@
...
+                c->execute("PRAGMA foreign_keys=OFF");
+
+                odb::core::transaction t{c->begin()};
+                odb::core::schema_catalog::create_schema(*db);
+                t.commit();
+
+                c->execute("PRAGMA foreign_keys=ON");
+            }
+        } catch (const odb::exception& e) {
+            throw std::runtime_error{std::string{"ODB report: failed ODB initialization: "} + e.what()};
```
cyenyxe: First part of the message could be a bit more user friendly, similar to sqlite_report.cpp: `"Can't initialize database: " + e.what()`.

src/vcf/odb_report.cpp
```diff
@@ -0,0 +1,150 @@
...
+        // possible recovery can be done here, ODB rollbacks automatically on error, and throws.
+        if (transaction.has_current()) {
+            transaction.commit();
+        }
+    }
+
+    // ReportWriter implementation
+    void OdbReportRW::write_error(Error &error)
+    {
+        error.get_severity() = Severity::ERROR;
```
cyenyxe: Better to define a setter than assign to the return value of a getter.

README.md
cyenyxe: Oops, you're right. Then I think the description is fine.

test/vcf/report_writer_test.cpp
```diff
@@ -19,6 +19,7 @@
 #include <fstream>
 
 #include <boost/filesystem.hpp>
+#include <vcf/odb_report.hpp>
```
cyenyxe: Include should be local to the project ("" instead of <>).

src/vcf/odb_report.cpp
```diff
@@ -0,0 +1,150 @@
...
+
+    size_t OdbReportRW::count(odb::query<ErrorCount> query)
+    {
+        ErrorCount count;
+        if (transaction.has_current()) {
+            throw std::logic_error{"There's another transaction active. You can only read if the changes were flushed"};
+        } else {
+            transaction.reset(db->begin());
+//        size_t count = db->execute("SELECT COUNT(*) FROM Error");
+            count = db->query_value<ErrorCount>(query);
```
cyenyxe: Is it possible to run read-only operations without a transaction?

src/vcf/odb_report.cpp
```diff
@@ -0,0 +1,150 @@
...
+        typedef odb::result<Error> result;
+
+        if (transaction.has_current()) {
+            throw std::logic_error{"There's another transaction active. You can only read if the changes were flushed"};
+        } else {
+            transaction.reset(db->begin());
+
+            result r{db->query<Error>(query + " ORDER BY " + odb::query<Error>::line)};
+
+            // TODO if result::iterator has operator*, operator!=, operator++, then use ranged for. not if we use shared_ptr
```
cyenyxe: Not sure I understand this comment. Does it still apply?

jmmut: doesn't apply anymore. First, the iterator was just accessed like `*i`, but now that we have to use `result::iterator::load()` I don't think we can use a ranged for. Comment removed in nex commit.

README.md
jmmut: fixed in next commit

inc/vcf/error.hpp
jmmut: fixed in next commit, moving to a LICENSE_ODB

src/main.cpp
jmmut: At the moment both have the same features. When the odb reporter outperforms the plain sqlite reporter, we will remove the sqlite one.

README.md
```diff
@@ -11,9 +11,17 @@ Please read the wiki for more details about checks already implemented.
 
 ## Dependencies
 
+### Boost
+
 The dependencies are the Boost library core, and its submodules: Boost.filesystem, Boost.program_options, Boost.regex and Boost.system.
 If you are using Ubuntu, the required packages names will be `libboost-dev`, `libboost-filesystem-dev`, `libboost-program-options-dev` and `libboost-regex-dev`.
 
+### ODB
```
cyenyxe: Mention that the SQLite library is a dependency and the required packages in Ubuntu are `libsqlite3-0` and `libsqlite3-dev`.

src/vcf/odb_report.cpp
jmmut: fixed in next commit

inc/vcf/error.hpp

src/vcf/odb_report.cpp
jmmut: fixed in next commit

test/vcf/report_writer_test.cpp
jmmut: fixed in next commit

src/vcf/odb_report.cpp
jmmut: haven't tried, but they use transactions for reading: http://www.codesynthesis.com/products/odb/doc/manual.xhtml#2.5

README.md
jmmut: fixed in next commit.

PR #41, opened 2016-08-25
cmake and readme improvements

jmmut:
The changes to CMakeLists.txt makes the project easier to debug, as it now allows to do out-of-source builds like this:
```
rm CMakeCache.txt
mkdir {build_release,build_debug}
cd build_release
cmake -G "Unix Makefiles" ..
make
cd ../build_debug
cmake -G "Unix Makefiles" -DCMAKE_BUILD_TYPE=DEBUG ..
make
```
By default, `CMAKE_BUILD_TYPE` is set to `RELWITHDEBINFO`. Be aware that the cmake cache may keep older values. Remove the cache or define the variables in the cmake call explicitly to be sure.
This structure is the recommended to have several build types at the same time, and the IDE takes advantage of not hardcoding `-O3` and `-g` flags.

Also, added debugulator command line examples.


cyenyxe:
Could the test files be also copied into the build folders? That would make the builds completely stand-alone.

CMakeLists.txt
```diff
@@ -5,10 +5,13 @@ set (vcf-validator_VERSION_MAJOR 0)
 set (vcf-validator_VERSION_MINOR 3)
 
 # no unknown pragmas: ODB compiler uses some pragmas that the regular compiler doesn't need
-set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -O3 -Wall -Wno-unknown-pragmas -g")
+set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -Wall -Wno-unknown-pragmas")
 # Needed to statically link ODB
 set (CMAKE_C_FLAGS " -DSQLITE_ENABLE_UNLOCK_NOTIFY")
 set (CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/bin)
+if(NOT CMAKE_BUILD_TYPE)
+    set (CMAKE_BUILD_TYPE  "RELWITHDEBINFO" CACHE STRING "Choose the type of build, options are: None(CMAKE_CXX_FLAGS or CMAKE_C_FLAGS used) Debug Release RelWithDebInfo MinSizeRel." FORCE)
```
cyenyxe: Maybe better to set the value to `RelWithDebInfo` to prevent any case-sensitivity issues?

README.md
```diff
@@ -42,6 +42,24 @@ In any case, the following binaries will be created in the `bin` subfolder:
 * `vcf_debugulator`: automatic fixing tool
 * `test_validator` and derivatives: testing correct behaviour of the tools listed above
 
+### ODB & Ragel build
```
cyenyxe: "Developers build"? That would prevent some end users to even look at this section without any further clarification, and if we have some other dependencies in the future the title is already generic.

README.md
```diff
@@ -71,8 +91,22 @@ The validation report can be exported in several ways with the `-r` / `--report`
...
+There are some simple errors that can be automatically fixed. The most common error is the presence of duplicate variants. The needed parameters are the original VCF and the report generated by a previous run of the vcf_validator with the option `-r database`.
+ 
+The fixed VCF will be written into the standard output, which you can redirect to a file, or use the `-o` / `--output` option and specify the desired file name.
+
+The logs about what the debugulator is doing will be written into the error output. The logs may be redirected to a log file `2>debugulator_log.txt` or completely discarded ` 2>/dev/null`
+
+## Examples
 
 Simple example: `vcf_validator -i /path/to/file.vcf -v v4.1`  
 Full example: `vcf_validator -i /path/to/file.vcf -v 4.1 -l stop -r database,stdout -o /path/to/output/folder/`
```
cyenyxe: @garysaunders detected an error in this example: it should state `-v v4.1` instead of `-v 4.1`.

README.md
```diff
@@ -71,8 +91,22 @@ The validation report can be exported in several ways with the `-r` / `--report`
...
 
+Debugulator example:
+```
+vcf_validator -i /path/to/file.vcf -v 4.1 -r database -o /path/to/write/report/
```
cyenyxe: This should state `-v v4.1` instead of `-v 4.1`.

CMakeLists.txt
jmmut: fixed in next commit

README.md
jmmut: fixed in next commit

README.md
jmmut: fixed in next commit

README.md
jmmut: fixed in next commit

PR #42, opened 2016-08-25
bug fix: debugulator: write all the file

jmmut:
the previous stopped copying the input file at the last error.

### DIFF
--- a/inc/util/stream_utils.hpp
+++ b/inc/util/stream_utils.hpp
@@ -45,6 +45,15 @@ namespace ebi
         return stream;
     }
 
+    template <typename Container>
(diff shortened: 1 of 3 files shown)


cyenyxe:
I just noticed that if an error can't be fixed, the line is not written either!

An easy way to test this is to run a 4.1 file with the validator in 4.2 mode. The fileformat declaration will be incorrect and the line won't be printed out.

Could more complex tests be added to check the number of lines in output files is correct?

jmmut:
I added as well the fix for the lines with error but no fix. I'm working on more sofisticated integration tests in the branch for INFO fixes.

PR #43, opened 2016-08-26
EVA-343 adding fix for invalid INFO fields

jmmut:
The first type of fix is just removing the wrong field.

If it was the only field in the INFO column, a "." is placed.

An INFO field qualifies as wrong if it has any syntax error (detected by the ragel part) or any semantic error (detected in record.cpp, as wrong cardinality or wrong type according to its definition in the meta section).


inc/vcf/error.hpp
```diff
@@ -269,8 +269,15 @@ namespace ebi
         using BodySectionError::BodySectionError;
```
cyenyxe: `BodySectionError` could contain a new attribute `num_column` to avoid using magic numbers in the fixer `visit` methods.

inc/vcf/fixer.hpp
```diff
@@ -108,9 +109,48 @@ namespace ebi
...
+
+            std::vector<std::string> columns;
+            std::string string_line{line->begin(), line->end()};
+            util::string_split(string_line, "\t", columns);
+            std::vector<std::string> infos;
+            util::string_split(columns[7], ";", infos);
+
+            for (size_t i = 0; i < 7; ++i) {
+                output << columns[i] << "\t";
+            }
```
cyenyxe: By using a column number as suggested [here](https://github.com/EBIvariation/vcf-validator/pull/43/files#r76741086), an auxiliary method `write_previous_fields` or similar could be extracted.

inc/vcf/fixer.hpp
```diff
@@ -108,9 +109,48 @@ namespace ebi
...
+        /**
+         * fix: remove the info field that caused the error
+         */
         virtual void visit(InfoBodyError &error) override
         {
-            ignored_errors++;
+            // TODO better log system, if any
+            std::cerr << "DEBUG: line " << error.get_line() << ": fixing invalid INFO field " << error.get_field()
+                      << std::endl;
+            int fixed = 0;
```
cyenyxe: `num_fixed_subfields` would be a more descriptive name.

inc/vcf/fixer.hpp
```diff
@@ -108,9 +109,48 @@ namespace ebi
...
+                    fixed++;
+                }
+            }
+
+            if (fixed == infos.size()) {
+                output << ".";
+            }
+            for (size_t i = 8; i < columns.size(); ++i) {
+                output << "\t" << columns[i];
+            }
```
cyenyxe: By using a column number as suggested [here](https://github.com/EBIvariation/vcf-validator/pull/43/files#r76741086), an auxiliary method `write_correct_fields` or similar could be extracted.

inc/vcf/fixer.hpp
```diff
@@ -108,9 +109,48 @@ namespace ebi
...
+                        output << ";";
+                    }
+                    output << infos[j];
+                } else {
+                    fixed++;
+                }
+            }
+
+            if (fixed == infos.size()) {
+                output << ".";
```
cyenyxe: Inline function `write_empty_field`?

inc/vcf/odb_report.hpp
```diff
@@ -18,13 +18,15 @@
...
 #include <odb/sqlite/query.hxx>
-#include "report_writer.hpp"
-#include "report_reader.hpp"
-
 #include <odb/schema-catalog.hxx>
 #include <odb/sqlite/database.hxx>
+
+#include "report_writer.hpp"
+#include "report_reader.hpp"
 #include "vcf/error-odb.hpp"
```
cyenyxe: Different inclusion reference, one with "vcf" prefix and others without.

inc/vcf/validator.hpp
```diff
@@ -173,7 +175,8 @@ namespace ebi
     using QuickValidator_v43 = ParserImpl_v43<QuickValidatorCfg>;
     using FullValidator_v43 = ParserImpl_v43<FullValidatorCfg>;
     using Reader_v43 = ParserImpl_v43<ReaderCfg>;
-    
+
+    std::unique_ptr<ebi::vcf::Parser> build_parser(std::string const & path, ValidationLevel level, ebi::vcf::Version version);
```
cyenyxe: `ebi::vcf` qualifiers shouldn't be necessary since they are already in the same namespace.

src/vcf/record.cpp
```diff
@@ -204,7 +204,9 @@ namespace ebi
                         check_field_cardinality(field.second, values, key_values["Number"], 2); // TODO Assumes ploidy=2
                         check_field_type(field.second, values, key_values["Type"]);
                     } catch (Error *ex) {
-                        throw new InfoBodyError{line, "Info " + key_values["ID"] + "=" + ex->get_raw_message()};
+                        throw new InfoBodyError{line,
+                                                "Info " + key_values["ID"] + "=" + ex->get_raw_message(),
```
cyenyxe: Would this construct a message like the following?

> Info AC = AC does not match the meta specification Number=A, expected 2 values

In that case, just forwarding the original exception could be enough.

src/main.cpp
```diff
@@ -174,80 +162,6 @@ namespace
         return outputs;
     }
 
-    std::unique_ptr<ebi::vcf::Parser> build_parser(std::string const & path, ValidationLevel level, ebi::vcf::Version version)
```
cyenyxe: Seems pretty sensible to move this into the validator namespace, since all the command line arguments have been already converted into objects of the business model.

test/vcf/debugulator_integration_test.cpp
```diff
@@ -52,79 +52,124 @@ namespace ebi
...
   {
-      auto path = boost::filesystem::path("test/input_files/v4.1/failed/failed_body_duplicated_000.vcf");
-      SECTION(path.string())
-      {
-          std::ifstream file{path.c_str()};
-          if (not file) {
-              throw std::runtime_error{"test file not found: " + path.string()};
+      std::ifstream file{path.c_str()};
+      if (not file) {
+          throw std::runtime_error{"test file not found: " + path.string()};
```
cyenyxe: The test should fail anyway if the file doesn't exist. If it doesn't, then it's an issue that this exception would hide.

test/vcf/debugulator_integration_test.cpp
```diff
@@ -52,79 +52,124 @@ namespace ebi
...
+      std::string debug_message;
+      long lines;
+      for (size_t i = 1; i <= 3; ++i) {
+          auto path = boost::filesystem::path("test/input_files/v4." + std::to_string(i) + "/failed/failed_body_duplicated_001.vcf");
+          SECTION(path.string()) {
+              std::tie(first_validation, validation_after_fixing, debug_message, lines) = fix_file(path, version);
+              CHECK_FALSE(first_validation);
+              INFO(debug_message);
+              CHECK(validation_after_fixing);
+              CHECK(lines == 6);
```
cyenyxe: These checks should be embedded into the testing function, otherwise it could run for a long time under incorrect assumptions (for instance, if `first_validation` is true is should fail straight away).

Splitting them in multiple functions would also improve readability.

test/vcf/debugulator_integration_test.cpp
```diff
@@ -52,79 +52,124 @@ namespace ebi
...
+              CHECK_FALSE(first_validation);
+              INFO(debug_message);
+              CHECK(validation_after_fixing);
+              CHECK(lines == 5);
           }
+          version = static_cast<ebi::vcf::Version>(i);
       }
   }
 }
-
```
cyenyxe: A test for files that don't require fixing would be also interesting: the debugulator should work fine under those circumstances too.

src/vcf/record.cpp
jmmut: no, for a vcf with this:
```
##INFO=<ID=MY,Number=G,Type=Integer,Description="INFO field to test Number=G">
...
... 	T	A	100	PASS	MY=1,3	
```
creates this message: `Line 5: Info MY=1,3 does not match the meta specification Number=G, expected 3 values`.
the inner message is `1,3 does not match the meta specification Number=G, expected 3 values`.
In my opinion, it is already quite clear .

cyenyxe: Yes, that is fine.

test/vcf/debugulator_integration_test.cpp
jmmut: The main reason why the checks are outside the helper function is that the CHECK* macros can only appear inside a TEST_CASE. I will try to improve it.

inc/vcf/fixer.hpp
jmmut: agree, fixed in next commit

inc/vcf/validator.hpp
jmmut: fixed in next commit

inc/vcf/odb_report.hpp
jmmut: fixed in next commit

inc/vcf/fixer.hpp
jmmut: done in next commit

inc/vcf/fixer.hpp
jmmut: same as below

inc/vcf/error.hpp
jmmut: not sure about this, a `DuplicationError` is also a body section error, but it doesn't refer to any specific column. I guess the Fixer class or its methods could define named constants for clarity

inc/vcf/fixer.hpp
jmmut: just like the column_number, I'd rather extract it as a constant string, because an empty field may be expressed differently across fields. however, in this case I don't mind adding a `write_empty_field` function, as it's more likely that `.` does apply as empty field in every column. So, function or const string?

src/main.cpp
jmmut: I'm a bit confused :P,  is this a comment stating that the change is OK?

test/vcf/debugulator_integration_test.cpp
jmmut: currently, the debugulator fails saying that the error report was empty. I thought it would be more "fail-fast"-y. Should we change that and allow an empty report to produce a copy of the VCF?

cyenyxe: It doesn't sound very useful to create a copy of the VCF when no errors are present, but it would be interesting to make the debugulator stop (more than fail) when the report is empty.

jmmut: Well, in fact, that is the current behaviour. I said "fails", but it just logs.

inc/vcf/fixer.hpp
```diff
@@ -108,9 +109,47 @@ namespace ebi
...
+            std::string string_line = {line->begin(), line->end()};
+
+            fix_column(info_column_index, string_line, "\t", [&](std::string &info_column) {
+                std::vector<std::string> info_subfields, keys_values;
+                util::string_split(info_column, ";", info_subfields);
+
+                for (size_t j = 0; j < info_subfields.size(); ++j) {
+                    util::string_split(info_subfields[j], "=", keys_values);
+                    if (keys_values[0] != error.get_field()) {
+                        if ((num_removed_subfields - j) != 0) {
```
cyenyxe: These 2 fields being of type `size_t`, subtracting them could cause overflow issues. Comparing one against the other should be fine.

jmmut: fixed in next commit

Issue #39, opened 2016-07-24
Fix INFO / FORMAT sub-field not matching meta


cyenyxe:
Fix the errors where an INFO or FORMAT sub-field doesn't match the definition in the meta section, most probably by removing it from the corresponding column.

PR #44, opened 2016-09-05
EVA-344 fix samples column not matching FORMAT and meta 

jmmut:
When the meta section has description about FORMAT fields, such as the Type (String, Integer, etc) or the Number (A, R, G, ., number) we can check that every sample value matches.

For any invalid field we find, we replace with "unknown value", commonly as a `.`.

For values whose number is more than 1, the `.` is repeated with commas (`,`) in between, except for the GT field, which is split by `/`. 

Added unit and integration tests.

### DIFF
--- a/inc/util/stream_utils.hpp
+++ b/inc/util/stream_utils.hpp
@@ -46,7 +46,7 @@ namespace ebi
     }
 
     template <typename Container>
-    std::ostream & writeline(std::ostream & stream, Container & container)
+    std::ostream & writeline(std::ostream & stream, const Container & container)
     {
         for (auto c : container) {
             stream << c;
@@ -61,19 +61,25 @@ namespace ebi
     }
 
     /**
-     * This is a generic method that prints a collection if it has the operations ".size()", ".begin()", ".end()"
+     * This is a generic method that prints a collection if its inner type is printable with "operator<<"
+     *
+     * @param os ostream to write into
+     * @param container must have he operations ".begin()" and ".end()"
      * defined and they provide iterators, and the inner type has an overloaded "operator<<"
+     * @param open_tag will be written before the first element (e.g. "[", if you want something like "[2,4]"), can be an empty string
+     * @param separator_tag will be written between elements (e.g. "," in "[2,4]"), can be empty string
+     * @param close_tag will be written after the last element (e.g. "]" in "[2,4]"), can be emtpy string
+     * @return the ostream passed, in the questionable case you want to do `print_containter(cout, ...) << "more messages"`;
      */
     template <typename T>
     std::ostream &print_container(std::ostream &os, const T &container,
                                   std::string open_tag, std::string separator_tag, std::string close_tag) {
-        size_t size = container.size();
         os << open_tag;
-        if (size > 0) {
-            auto it = container.begin();
+        auto it = container.begin();
+        auto end = container.end();
+        if (it != end) {
             os << *it;
-            it++;
-            for (; it != container.end(); ++it) {
+            for (++it; it != container.end(); ++it) {
                 os << separator_tag << *it;
             }
         }
@@ -82,8 +88,10 @@ namespace ebi
     }
 
     /**
-     * This is a generic method that prints a collection if it has the operations ".size()", ".begin()", ".end()"
+     * This is a generic method that prints a collection if it has the operations ".begin()" and ".end()"
      * defined and they provide iterators, and the inner type has an overloaded "operator<<"
+     *
+     * provides the default open/close/separator tags to print collections like this: [2,4]
      */
     template <typename T>
     std::ostream &print_container(std::ostream &os, const T &container) {
@@ -99,7 +107,7 @@ namespace ebi
     template <typename K, typename V>
     std::ostream &operator<<(std::ostream &os, const std::map<K, V> &container)
     {
-        return print_container<std::map<std::string, std::string>>(os, container);
+        return print_container<std::map<K, V>>(os, container);
     }
   }
 }
--- a/inc/util/stream_utils.hpp
+++ b/inc/util/stream_utils.hpp
@@ -59,24 +59,35 @@ namespace ebi
     {
         return os << "{" << container.first << ", " << container.second << "}";
     }
+
     /**
-     * This is a generic method that prints a collection if it has the operations ".size()", ".begin()", ".end()" 
+     * This is a generic method that prints a collection if it has the operations ".size()", ".begin()", ".end()"
      * defined and they provide iterators, and the inner type has an overloaded "operator<<"
      */
     template <typename T>
-    std::ostream &print_container(std::ostream &os, const T &container) {
+    std::ostream &print_container(std::ostream &os, const T &container,
+                                  std::string open_tag, std::string separator_tag, std::string close_tag) {
         size_t size = container.size();
-        os << "[";
+        os << open_tag;
         if (size > 0) {
             auto it = container.begin();
             os << *it;
             it++;
             for (; it != container.end(); ++it) {
-                os << ", " << *it;
+                os << separator_tag << *it;
             }
         }
-        os << "]";
+        os << close_tag;
         return os;
+    }
+
+    /**
+     * This is a generic method that prints a collection if it has the operations ".size()", ".begin()", ".end()"
+     * defined and they provide iterators, and the inner type has an overloaded "operator<<"
+     */
+    template <typename T>
+    std::ostream &print_container(std::ostream &os, const T &container) {
+        return print_container(os, container, "[", ", ", "]");
     }
     
     // the next functions allow using the previous print_container in a easier syntax
--- a/inc/vcf/error.hpp
+++ b/inc/vcf/error.hpp
@@ -51,24 +51,24 @@ namespace ebi
 
     enum class Severity { WARNING, ERROR };
 
-    class Error;
-    class MetaSectionError;
-    class HeaderSectionError;
-    class BodySectionError;
-    class FileformatError;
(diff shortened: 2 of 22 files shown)


jmmut:
something happened to clang packages. gonna do some research

jmmut:
also, closes #39

inc/util/stream_utils.hpp
```diff
@@ -59,25 +59,36 @@ namespace ebi
...
             }
         }
-        os << "]";
+        os << close_tag;
         return os;
     }
+
+    /**
+     * This is a generic method that prints a collection if it has the operations ".size()", ".begin()", ".end()"
+     * defined and they provide iterators, and the inner type has an overloaded "operator<<"
```
cyenyxe: Describe what are the default tags.

inc/vcf/fixer.hpp
```diff
@@ -156,11 +157,68 @@ namespace ebi
...
+         * - get the index in the FORMAT column of error.get_field(), e.g. in GT:AC:AN, the index of AC is 1.
+         * - for each sample column
+         *      - put a "." instead of the value in the index-th position of ":" e.g. 0/0:3:4 becomes 0/0:.:4 for index = 1
+         * @param error
+         */
         virtual void visit(SamplesBodyError &error) override
         {
-            util::writeline(output, *line);
-            ignored_errors++;
+            if (error.get_field().empty()) {
```
cyenyxe: I think that for now it could make sense to make `field` mandatory. If we were to solve other type of `SamplesBodyError`, then having a richer type hierarchy would be best than mixing different fixes in the same method.

inc/vcf/fixer.hpp
```diff
@@ -156,11 +157,68 @@ namespace ebi
...
+                util::writeline(output, *line);
+                ignored_errors++;
+            } else {
+                std::cerr << "DEBUG: line " << error.get_line() << ": fixing invalid sample field " << error.get_field()
+                          << std::endl;
+
+                const size_t format_column = 8;
```
cyenyxe: `format_column_index` for consistency with the INFO visitor method.

inc/vcf/fixer.hpp
jmmut: I did not make it mandatory because the next errors are not related to any specific field:
- "the format lists a different number of fields than the sample column has"
- "The number of samples must match those listed in the header line"
- general syntax errors detected in ragel

inc/vcf/fixer.hpp
```diff
@@ -156,11 +157,68 @@ namespace ebi
...
+                        return;
+                    }
+                    output << *first;   // write the FORMAT column
+
+                    // if the field is GT, the values must use "/", as in "./."
+                    const std::string subfield_separator = subfield_index == 0? "/" : ",";
+
+                    // error.number should be -1 if the cardinality is unknown, and any positive number otherwise
+                    // so, if unknown, put 1, so that a single "." is written
+                    size_t number = error.get_number() <= -1? 1 : static_cast<size_t>(error.get_number());
```
cyenyxe: Could these `number` variables be renamed to something like `field_cardinality`? It is not clear at first sight what their purpose is.

inc/vcf/fixer.hpp
```diff
@@ -156,11 +157,68 @@ namespace ebi
...
             ignored_errors++;
         }
+
+        /**
+         * explanation of the fix:
+         * - in error.get_field there will be the field name as it appears in the FORMAT column, e.g. "GT".
+         * - don't do anything if error.get_field() is empty
+         * - get the index in the FORMAT column of error.get_field(), e.g. in GT:AC:AN, the index of AC is 1.
+         * - for each sample column
+         *      - put a "." instead of the value in the index-th position of ":" e.g. 0/0:3:4 becomes 0/0:.:4 for index = 1
```
cyenyxe: I was thinking, since the field will be made missing for every sample, wouldn't it make more sense to remove it from the FORMAT and samples columns?

inc/vcf/fixer.hpp
```diff
@@ -174,29 +232,108 @@ namespace ebi
         }
 
       protected:
+
+        size_t split_and_find(std::string line, std::string separator, std::string value) {
```
cyenyxe: Seems like a perfect candidate to be in the `util` namespace :)

inc/vcf/fixer.hpp
```diff
@@ -174,29 +232,108 @@ namespace ebi
...
+                        long column_index_last,
+                        const std::string &line,
+                        std::string separator,
+                        std::function<void(std::vector<std::string>::iterator begin,
+                                           std::vector<std::string>::iterator end)> fix_function) {
+
             std::vector<std::string> columns;
             util::string_split(line, separator.c_str(), columns);
+            std::string eol;
+            if (columns.back().back() == '\n') {
```
cyenyxe: The last version of the specification requires to support both `\n` and `\r\n`.

inc/util/stream_utils.hpp
jmmut: fixed in next commit

inc/vcf/fixer.hpp
jmmut: fixed in next commit

inc/vcf/fixer.hpp
jmmut: fixed in next commit

inc/vcf/fixer.hpp
jmmut: in next commit I will use const references and there will be a documentation comment.

Some context on usage: if we have `"GT:AC:AN    0/0:3:4"` and have to remove the AC field, we need to know that it's in position 1 for later when we split the sample column: `["0/0", "3", "4"]`. 

`split_and_find(format_column, ...)` returns that position 1 from the FORMAT column, or `format_column.npos` if it was not found (just as in `string::find` methods).
We cannot return iterators because those will be pointing to the split `vector<string>` local to `split_and_find`.

inc/vcf/fixer.hpp
jmmut: fixed in next commit

inc/vcf/fixer.hpp
jmmut: fixed in next commit

inc/vcf/fixer.hpp
```diff
@@ -174,29 +223,272 @@ namespace ebi
...
+         */
+        void fix_format_gt(std::vector<std::string>::iterator first,
+                           std::vector<std::string>::iterator last,
+                           SamplesFieldBodyError &error)
+        {
+            const std::string field_separator = ":";
+            size_t gt_column_index = 0;
+            // if the field is GT, the values must use "/", as in "./."
+            const std::string subfield_separator = "/";
+            const char empty_subfield = '.';
```
cyenyxe: The 3 fields above could be char instead of string.

inc/vcf/error.hpp
```diff
@@ -299,6 +301,29 @@ namespace ebi
...
     #pragma db object
+    class SamplesFieldBodyError : public BodySectionError
+    {
+      private:
+        friend class odb::access;
+        SamplesFieldBodyError() {}  // necessary for ODB
+      public:
+        SamplesFieldBodyError(size_t line,
+                         const std::string &message,
+                         std::string field,
```
cyenyxe: Why not `const std::string &` ?

inc/vcf/fixer.hpp
```diff
@@ -161,6 +149,67 @@ namespace ebi
...
+         * - if there were no sample columns, skip fix (copy as-is)
+         *
+         * complete example:
+         * field "GT", FORMAT: "GT:AC:AN", sample: "0/0:3,5:4"; becomes "./.:3,5:4". (index = 0, cardinality = 2)
+         * field "AC", FORMAT: "GT:AC:AN", sample: "0/0:3,5:4"; becomes "GT:AN   0/0:4". (index = 1)
+         * @param error
+         */
+        virtual void visit(SamplesFieldBodyError &error) override
+        {
+            if (error.get_field().empty()) {
```
cyenyxe: I'm not sure we should allow to build this type of error using an empty value... Could it be better to implement a check in the constructor and get rid of this if-else?

inc/vcf/fixer.hpp
```diff
@@ -117,33 +118,20 @@ namespace ebi
...
-                        if (num_removed_subfields != j) {
-                            // if this is not the first field we don't remove, then write the separator
-                            output << ";";
-                        }
-                        output << info_subfields[j];
-                    } else {
-                        num_removed_subfields++;
-                    }
-                }
+            auto remove_info_field_if = [&](std::string &info_subfield, size_t index) -> bool {
```
cyenyxe: Could this be renamed to `condition_to_remove_info_field`? I didn't understand what it did until I saw the `remove_column` definition.

inc/vcf/fixer.hpp
jmmut: I saw that this check is redundant, because it is handled as if it were an invalid field, so the `if` can be safely, and I added an exception in the `SamplesFieldBodyError` constructor in case it is empty. We have the SamplesBodyError for more undefined errors in the samples columns

inc/vcf/fixer.hpp
jmmut: fixed in next commit

inc/vcf/error.hpp

inc/vcf/fixer.hpp
jmmut: our string_split allows several separators and and print_container uses strings for separators. We could use chars, but then we would need conversions to string later.
I would let this as-is.

PR #46, opened 2016-09-21
EVA-407 removed some useless shared_ptrs from PasingState and Parsers

jmmut:
in parsing state:
- `source` is still a shared pointer because it has to be shared with each record
- `record` is not a vector anymore, as there's only one at each moment. unique pointer now.
- `errors` and `warnings` are not pointers to vectors anymore, but plain vectors. (vectors to unique pointers to `Error`s)

### DIFF
--- a/inc/vcf/parsing_state.hpp
+++ b/inc/vcf/parsing_state.hpp
@@ -41,22 +41,21 @@ namespace ebi
         bool m_is_valid;
         
         std::shared_ptr<Source> source;
-        std::shared_ptr<std::vector<Record>> records;
-        std::shared_ptr<std::vector<std::unique_ptr<Error>>> errors;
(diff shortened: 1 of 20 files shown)


inc/vcf/parsing_state.hpp
```diff
@@ -41,22 +41,21 @@ namespace ebi
...
+        void set_version(Version version);
         
-        void add_meta(MetaEntry const & meta) const;
-        
-        void add_record(Record const & record);
-        void clear_records() const;
+        void add_meta(MetaEntry const & meta);
+
+        void add_record(std::unique_ptr<Record> record);
+        void clear_records();
```
cyenyxe: Rename the first to `set_record` and maybe remove the second one? Otherwise it could be named `unset_record` instead of `clear_records`.

jmmut: fixed in next commit

PR #48, opened 2016-10-02
EVA-407 Fixed high memory usage caused by 17ef05e

cyenyxe:
Replaced the `unique_ptr<Record>` attribute of `ParsingState` with a plain object. This made necessary to implement the default constructor for the `Record` class.

Some small refactoring to the `ParsingState` class, grouping all the attribute "clearing" methods into one.


jmmut:
the changes in this PR were included in other branches in a more organized way

PR #51, opened 2016-10-10
EVA-452 Docker image for static builds

cyenyxe:
The `docker` subfolder contains Dockerfile and an entry point for creating a container to build static binaries easily. The reason not to distribute directly a binary are:

* Single container not suitable for development and production: development would require the last local commit, while production would match a specific (remote) tag.
* Non trivial files mapping from host to guest. To run a validation, the VCF file would need to map to a specific path in the guest, as in: `docker run -v /host/path/to/file.vcf:/tmp/file.vcf ebivariation/vcf-validator`.
* Docker runs as root, which would make not possible to deploy it in restricted cluster environments.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,16 +11,16 @@ Please read the wiki for more details about checks already implemented.
 
 ## Build
 
-The easiest way to build vcf-validator is using the Docker image provided with the source code. This will create a binary that can be run in any Linux machine.
+The easiest way to build vcf-validator is using the Docker image provided with the source code. This will create an executable that can be run in any Linux machine.
 
 1. Install and configure Docker following [their tutorial](https://docs.docker.com/engine/getstarted/).
 2. Create the Docker image:
-    1. Clone this Git repository
-    2. Move to the folder the code was downloaded to
-    3. Run `docker build -t my-name/vcf-validator docker/`. You can replace `my-name` with your name of choice.
-3. Build the application running `docker run -v ${PWD}:/tmp my-name/vcf-validator`. Replace `my-name` with the name used in the previous step.
+    1. Clone this Git repository: `git clone https://github.com/EBIvariation/vcf-validator.git`
+    2. Move to the folder the code was downloaded to: `cd vcf-validator`
+    3. Build the image: `docker build -t ebivariation/vcf-validator docker/`. Please replace `ebivariation` with your user account if you plan to push this image to [Docker Hub](https://hub.docker.com).
+3. Build the executable running `docker run -v ${PWD}:/tmp ebivariation/vcf-validator`. Again, replace `ebivariation` with your user name if necessary.
 
-The following binaries will be created in the `build/bin` subfolder:
+The following executables will be created in the `build/bin` subfolder:
 
 * `vcf_validator`: validation tool
 * `vcf_debugulator`: automatic fixing tool
--- a/README.md
+++ b/README.md
@@ -9,62 +9,22 @@ It includes all the checks from the vcftools suite, and some more that involve l
 
 Please read the wiki for more details about checks already implemented.
 
-## Dependencies
-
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -9,63 +9,23 @@ It includes all the checks from the vcftools suite, and some more that involve l
...
-* Clang 3.5 to 3.7
-* GCC 4.8 to 5.0
+The easiest way to build vcf-validator is using the Docker image provided with the source code. This will create a binary that can be run in any Linux machine.
 
-In order to create the build scripts, please run `cmake` with your preferred generator. For instance, `cmake -G "Unix Makefiles"` will create Makefiles, and to build the binaries, you will need to run `make`. If the ODB libraries were not found during the build, please run `sudo updatedb && sudo ldconfig`.
+1. Install and configure Docker following [their tutorial](https://docs.docker.com/engine/getstarted/).
+2. Create the Docker image:
+    1. Clone this Git repository
+    2. Move to the folder the code was downloaded to
+    3. Run `docker build -t my-name/vcf-validator docker/`. You can replace `my-name` with your name of choice.
```
cyenyxe: With actual example commands, you mean adding them explicitly to 1 and 2? I'm fine with that.

Explaining what `my-name` is would require a lot of text regarding what are Docker repositories... I can change the clarification to something like "You can replace `my-name` with any name that could identify you uniquely" and maybe add a link for those interested or in doubt. Otherwise suggestions are welcome.

The input VCF paths in the examples will obviously differ, but I hope a bioinformatician will be able to extrapolate their meaning :stuck_out_tongue: I don't understand the last question "Is this my-name equally unguessable?"

docker/Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+
+# Ubuntu-native ODB packages don't seem to work due to a broken ABI, see this thread: http://codesynthesis.com/pipermail/odb-users/2016-May/003277.html
+RUN wget http://www.codesynthesis.com/download/odb/2.4/odb_2.4.0-1_amd64.deb -O /opt/odb_2.4.0-1_amd64.deb
+RUN wget http://codesynthesis.com/download/odb/2.4/libodb-2.4.0.tar.bz2 -O /opt/libodb.tar.bz2
+RUN wget http://codesynthesis.com/download/odb/2.4/libodb-sqlite-2.4.0.tar.bz2 -O /opt/libodb-sqlite.tar.bz2
+
+# ODB compiler
+RUN cd /opt && tar jxvf libodb.tar.bz2 && tar jxvf libodb-sqlite.tar.bz2 && dpkg -i odb_2.4.0-1_amd64.deb
+RUN cd /opt/libodb-2.4.0 && ./configure && make && make install
+RUN cd /opt/libodb-sqlite-2.4.0 && ./configure --with-libodb=../libodb-2.4.0 && make && make install
```
cyenyxe: I'm not sure, parallel `make` runs could be incorrect if the folder hierarchy is not properly managed, but by using `configure` it should be fine. I will give it a try.

docker/entrypoint.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+
+NPROCS=`grep processor </proc/cpuinfo | wc -l`
+
+# Create build directory
+mkdir -p /tmp/build
+cd /tmp/build
+
+# Generate last version of parser sources from Ragel descriptors
+for i in 1 2 3 ; do ragel -G2 -o ../inc/vcf/validator_detail_v4${i}.hpp ../src/vcf/vcf_v4${i}.ragel ; done
```
cyenyxe: I will change the previous command to `cd /tmp` so the `..` are not necessary, and add `cd /tmp/build` after the loop, to ensure the binaries are built inside that subfolder.

PR #52, opened 2016-10-12
EVA-470 Summary report independent from checks

jmmut:
Allow filtering superfluous warnings when writing to standard output.

The checks are always done and always raised, but only reported in non-summary reports.
The previous behaviour was that if the error was already raised, it was not raised for any report type.

### DIFF
--- a/inc/vcf/error.hpp
+++ b/inc/vcf/error.hpp
@@ -183,8 +183,11 @@ namespace ebi
         friend class odb::access;
         NoMetaDefinitionError() {}
       public:
-        NoMetaDefinitionError(size_t line, std::string message, std::string column, std::string field)
-                : Error{line, field  + " from column " + column + " was not defined in meta section"} { }
+        NoMetaDefinitionError(size_t line,
+                              const std::string &message,
+                              const std::string &column,
+                              const std::string &field)
+                : Error{line, message}, column{column}, field{field} {}
         virtual ErrorCode get_code() const override { return ErrorCode::body_section; }
         virtual void apply_visitor(ErrorVisitor &visitor) override { visitor.visit(*this); }
         std::string column;
--- a/inc/vcf/error.hpp
+++ b/inc/vcf/error.hpp
@@ -179,9 +179,6 @@ namespace ebi
     #pragma db object
     struct NoMetaDefinitionError : public Error
     {
-      private:
-        friend class odb::access;
-        NoMetaDefinitionError() {}
       public:
         NoMetaDefinitionError(size_t line,
                               const std::string &message,
@@ -192,6 +189,9 @@ namespace ebi
         virtual void apply_visitor(ErrorVisitor &visitor) override { visitor.visit(*this); }
         std::string column;
         std::string field;
+      private:
+        friend class odb::access;
+        NoMetaDefinitionError() {}
     };
 
     // inheritance siblings about detailed errors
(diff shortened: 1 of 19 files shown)


inc/vcf/error.hpp
```diff
@@ -172,6 +176,21 @@ namespace ebi
...
     };
 
+    #pragma db object
+    struct NoMetaDefinitionError : public Error
+    {
+      private:
+        friend class odb::access;
+        NoMetaDefinitionError() {}
+      public:
+        NoMetaDefinitionError(size_t line, std::string message, std::string column, std::string field)
```
cyenyxe: Other classes in the hierarchy use `const std::string &` as string arguments for the constructor.

inc/vcf/summary_report_writer.hpp
```diff
@@ -0,0 +1,107 @@
...
+        bool skip;
+    };
+
+
+    class SummaryReportWriter : public ReportWriter
+    {
+      public:
+        virtual void write_error(Error &error)
+        {
+          std::cout << error.what() << std::endl;
```
cyenyxe: This will conflict with the plain text report, because they both write to stdout: https://github.com/jmmut/vcf-validator/blob/76ad342fc2cb16af51aa2d91fc464a7bd1354060/inc/vcf/report_writer.hpp#L42

jmmut: complete report removed, so now only the summary report writes to stdout

src/vcf/error.cpp
```diff
@@ -33,6 +33,8 @@ namespace ebi
             return std::shared_ptr<Error>(new HeaderSectionError{line, message});
         case ErrorCode::body_section:
             return std::shared_ptr<Error>(new BodySectionError{line, message});
+        case ErrorCode::no_meta_definition:
```
jmmut: yes

inc/vcf/summary_report_writer.hpp
```diff
@@ -0,0 +1,134 @@
...
+        virtual void visit(SamplesBodyError &error) {}
+        virtual void visit(SamplesFieldBodyError &error) {}
+        virtual void visit(NormalizationError &error) {}
+        virtual void visit(DuplicationError &error) {}
+
+      private:
+        bool is_already_reported(std::string const &meta_type, std::string const &id) const
+        {
+            typedef std::multimap<std::string,std::string>::const_iterator iter;
+            std::pair<iter, iter> range = undefined_metadata.equal_range(meta_type);
```
jmmut: fixed in next commit

inc/vcf/summary_report_writer.hpp
```diff
@@ -0,0 +1,134 @@
...
+
+
+
+    /**
+     * Implements a ReportWriter that writes to stdout, but small warnings are written only once.
+     */
+    class SummaryReportWriter : public ReportWriter
+    {
+      public:
+        SummaryReportWriter(std::ostream &out) : out(out) {}
```
jmmut: fixed in next commit

inc/vcf/error.hpp
```diff
@@ -172,6 +176,24 @@ namespace ebi
         virtual void apply_visitor(ErrorVisitor &visitor) override { visitor.visit(*this); }
     };
 
+    #pragma db object
+    struct NoMetaDefinitionError : public Error
+    {
+      private:
```

src/main.cpp
```diff
@@ -196,7 +197,7 @@ namespace
                 }
                 outputs.emplace_back(new ebi::vcf::OdbReportRW(db_filename));
             } else if (out == "stdout") {
```
cyenyxe: Agreed to think a bit more about this and make a decision in the short term.

inc/vcf/summary_report_writer.hpp
jmmut: reverted, it breaks gcc 4.8 compilation

PR #54, opened 2016-10-25
EVA-499 remove SqliteReport and ErrorCodes 

jmmut:
both things are split in two commits, just in case we don't want the removal of ErrorCode.


CMakeLists.txt
```diff
@@ -131,15 +125,13 @@ if (BUILD_STATIC)
         ${Boost_LIBRARIES}
         ${ODB_PATH}/libodb-sqlite.a
         ${ODB_PATH}/libodb.a
-        sqlite3
```
cyenyxe: Has the Docker build being tested after removing this? If it passes, then line 10 in `CMakeLists.txt` should be no longer necessary:

> set (CMAKE_C_FLAGS " -DSQLITE_ENABLE_UNLOCK_NOTIFY -DSQLITE_OMIT_LOAD_EXTENSION")

jmmut: After seeing this, I noticed the static build was failing. Since we no longer compile the sqlite, the libsqlite.a in the system has `dl` enabled, so the warnings are raised: 
> /usr/lib/x86_64-linux-gnu/libsqlite3.a(sqlite3.o): In function `unixDlOpen':
> (.text+0x9419): warning: Using 'dlopen' in statically linked applications requires at runtime the shared libraries from the glibc version used for linking

Having that warning in mind, the static build and the docker build work, and I think this is the closer we can get without compiling sqlite ourselves

jmmut:
Well, I updated the PR to keep sqlite but remove `sqlite_report` and `ErrorCode`. This allow correct builds (regular, static and docker-based)

PR #59, opened 2016-11-24
fix bug: string split on empty strings

jmmut:
Symptom:
- different behaviour parsing info flags, depending on the version of the c++
standard library used.

Explanation of the effect chain:
- in 5.4 c++ library version (unlike in 4.8), the memory seems not to be
erased, on some string operations.
- using 5.4, sometimes in our program appear empty strings like "\0ns,type\0"
of length 0.
- the previous version of string_split (when the input string was empty) was
using past-the-end bytes.
- the split returned a vector of size 0 or 1 depending on the library version.
- the same vcf could be valid or invalid depending on the library version.

### DIFF
--- a/inc/util/string_utils.hpp
+++ b/inc/util/string_utils.hpp
@@ -25,23 +25,33 @@ namespace ebi
 {
   namespace util
   {
+    /**
+     * Splits `s` using `delims` as separator and fills the container `ret` with the parts.
+     * An empty string results in an empty container `ret`.
+     * @param s input string to split
+     * @param delims any character here acts as a separator
+     * @param ret return by reference the container filled with the string split.
+     */
     template<typename C>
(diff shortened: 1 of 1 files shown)


inc/util/string_utils.hpp
```diff
@@ -28,23 +28,25 @@ namespace ebi
     template<typename C>
     void string_split(std::string const & s, char const * delims, C & ret)
     {
-        C output;
-        char const* p = s.c_str();
-        char const* q = strpbrk(p+1, delims);
+        if (s.size() > 0) {
```
cyenyxe: Making this check the other way around would minimize the indentation (and the diff):

```
if (s.size() == 0) {
    return;
}

C output;
...
```

jmmut: Just thought that that extra return was not adding clarity.
Amended.

Issue #35, opened 2016-07-20
Do not allow repeated values in single line


cyenyxe:
Version 4.3 of the specification describes the ID as:

>  No identifier should be present in more than one data record. String, no white-space or semi-colons permitted, duplicate values not allowed.

The FILTER, INFO and FORMAT columns are described similarly.

This means that if duplicated values are found in a single line, an error must be thrown. Across multiple lines it could be a warning.

cyenyxe:
Fixed by #63, #64 and #76.

PR #63, opened 2017-04-07
Added single-line duplicate values check for ID

Anishka0107:
This PR aims to partly fix the issue https://github.com/EBIvariation/vcf-validator/issues/35
According to VCF 4.3 specifications, ID cannot have duplicate values, so added a new check for duplicate values. If any duplicated IDs are found in a single line (semicolon separated list), then an error is thrown and file is declared as invalid.

### DIFF
--- a/src/vcf/record.cpp
+++ b/src/vcf/record.cpp
@@ -122,8 +122,8 @@ namespace ebi
         
         std::map<std::string, int> counter;
         for (auto & id : ids) {
-            counter[*id]++;
-            if (counter[*id] >= 2) {
+            counter[id]++;
+            if (counter[id] >= 2) {
                 throw new IdBodyError{line, "ID must not have duplicate values"};
             }
         }
--- a/src/vcf/record.cpp
+++ b/src/vcf/record.cpp
@@ -119,6 +119,14 @@ namespace ebi
                 throw new IdBodyError{line, "ID must not contain semicolons or whitespaces"};
             }
         }
+        
+        std::map<std::string, int> counter;
+        for (auto id = ids.begin(); id != ids.end(); ++id) {
+            counter[*id]++;
+            if (counter[*id] >= 2) {
+                throw new IdBodyError{line, "ID must not have duplicate values"};   
+            }
+        }
     }
 
     void Record::check_alternate_alleles() const
--- a/src/vcf/record.cpp
+++ b/src/vcf/record.cpp
@@ -121,10 +121,10 @@ namespace ebi
         }
         
         std::map<std::string, int> counter;
-        for (auto id = ids.begin(); id != ids.end(); ++id) {
+        for (auto & id : ids) {
             counter[*id]++;
             if (counter[*id] >= 2) {
-                throw new IdBodyError{line, "ID must not have duplicate values"};   
+                throw new IdBodyError{line, "ID must not have duplicate values"};
             }
         }
     }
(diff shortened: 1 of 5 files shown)


Anishka0107:
@jmmut Thanks for the review sir. 
I have made a few changes as you suggested, but now the Travis build is failing and I am unable to understand why that is happening. The test was successful locally.

jorizci:
It's failing because...

/home/travis/build/EBIvariation/vcf-validator/src/vcf/record.cpp:125:21: error: indirection requires pointer operand ('const std::basic_string<char>' invalid)
            counter[*id]++;
                    ^~~
/home/travis/build/EBIvariation/vcf-validator/src/vcf/record.cpp:126:25: error: indirection requires pointer operand ('const std::basic_string<char>' invalid)
            if (counter[*id] >= 2) {

jmmut:
oh, yes, id is not a pointer anymore, so removing the `*` fixes it. If your tests passed, maybe you forgot to include in the commit all the changes?.

jorizci:
@Anishka0107 Just a little tip for any git-related project. If you have a situation where part of the changes haven't been commited (It just happens) you can use the commit ammend functionality (In command line you do a git add files where missing changes are and then use the git commit --amend and do a push force to your branch) that way the historic stays cleaner and you can avoid having a lot of commits with "missing ..." Anyway just wait for @jmmut to check everything , approve the pr and merge the code the repository.

jmmut:
Don't worry, I guessed that because I've done that dozens of times. I actually forgot to ask you this morning to add a test for the new feature, sorry. 

Under the  test/vcf folder we have the tests. Some of them are unitary (for instance record_test.cpp) and other are integration tests (such as parser_v43_test.cpp).  

For this PR just add another `SECTION` clause right after this one https://github.com/EBIvariation/vcf-validator/blob/develop/test/vcf/record_test.cpp#L166, which is very similar, and change the IDs so that it checks the new feature. We are using Catch ([tutorial](https://github.com/philsquared/Catch/blob/master/docs/tutorial.md), [assertions](https://github.com/philsquared/Catch/blob/master/docs/assertions.md))

Anishka0107:
@jorizci That's really cool. Thanks for the help:)
@jmmut Working on it sir.

Anishka0107:
@jmmut I have added the test.

cyenyxe:
There is only one more test to implement, in order to check that the whole application marks as invalid a VCF file with this kind of error. If you go to `test/input_files/` you will find one subfolder per version of the specification, and inside that one, a `failed` subfolder containing test VCF files.

These are named according to the section and field they validate (meta_alt, body_chrom, **body_id**). In this case you need to do the following:

1. Create a new file `failed_body_id_003.vcf` inside each `test/input_files/v4.x/failed` folder. You can use another file in the same "series" as reference
2. Make the ID field have a duplicate value
3. Indicate what is wrong in the `CauseOfFailure` message
4. The test suite will take care of the rest

Anishka0107:
@cyenyxe I have added 3 test files with duplicate IDs.

cyenyxe:
Looks great, thanks! Other than the comment below I think this is ready to merge 😄 

Taking previous experience into account, most of the times the ID field contains only one value. Could you please add a condition to avoid creating the std::map and running the validation in that case?

Anishka0107:
@cyenyxe Added the condition ma'am.
For it to look simple, I am checking for the size of the vector, is that okay? Or should I use count algorithm of STL or something else.

cyenyxe:
I think just checking the size is the best / simplest solution.

PR #64, opened 2017-04-10
Added single-line duplicate fields check for FORMAT

Anishka0107:
According to VCF 4.3 specifications, FORMAT cannot have duplicate fields, so added a new check for duplicate fields. If any duplicated FORMATs are found in a single line (colon separated list), then an error is thrown and file is declared as invalid.


src/vcf/record.cpp
```diff
@@ -236,6 +236,16 @@ namespace ebi
         if (std::find(format.begin(), format.end(), "GT") != format.end() && format[0] != "GT") {
             throw new FormatBodyError{line, "GT must be the first field in the FORMAT column"};
         }
+
+        if (format.size() > 1) {
```
jorizci: when format == 0 the function has already returned before this code, unless size can be negative it is always gonna be >1 right?

cyenyxe: It could be also 1, in which case the loop is not necessary.

Anishka0107:
@cyenyxe Sorry I forgot the test, I have added one in a new commit. Could you please take a look?

cyenyxe:
We have read through the specification documents again and it seems that non-duplicated checks for FORMAT, INFO and FILTER only apply to version 4.3.

Could you please add that condition to the validation you just implemented? The version can be obtained from the `source` attribute in the `Record` object. Test cases and associated files will also need to be reorganized.

Anishka0107:
This is exactly what I was wondering about. I'll work on it. Isn't it needed for ID ma'am?

cyenyxe:
You are right, it is also needed for the ID. The paragraph describing that field is more ambiguous than the others...

Anishka0107:
@cyenyxe @jmmut I have a doubt. The https://github.com/EBIvariation/vcf-validator/blob/develop/src/vcf/record.cpp file only has a test for version 4.1
So for implementing this new test in 4.3, I guess I will have to add another source. Am I guessing it right?

Anishka0107:
Added another commit with the checks being removed from v4.1 and v4.2
Now it would only be tested for v4.3

PR #65, opened 2017-04-11
EVA-720 Detect VCF version automatically

jmmut:
Basically, remove the `-v` / `--version` parameter. 

To achieve this (even reading from stdin) we have to read the first line, choose a parser version, and send the first line to the parser to be processed again.

A better design would be a further refactor that joins the 3 ragel machines (one per VCF version) in a single machine, that branches into 3 sub-machines after the first line. To keep the version grammars independent (and extract common actions and basic machines) we can use the "include" feature from ragel, following the structure in [this example repo](https://github.com/jmmut/ragel-includes-test)

### DIFF
--- a/src/vcf/validator.cpp
+++ b/src/vcf/validator.cpp
@@ -20,6 +20,12 @@ namespace ebi
 {
   namespace vcf
   {
+    bool validate(const std::vector<char> &firstLine,
+                           std::istream &input,
+                           ebi::vcf::Parser &validator,
+                           std::vector<std::unique_ptr<ebi::vcf::ReportWriter>> &outputs);
+
+    void write_errors(const Parser &validator, const std::vector<std::unique_ptr<ReportWriter>> &outputs);
 
     ParserImpl::ParserImpl(std::shared_ptr<Source> source)
             : ParsingState{source}
@@ -127,35 +133,46 @@ namespace ebi
     {
         std::vector<char> line;
         ebi::util::readline(input, line);
-        ebi::vcf::Version version = extract_version(line);
+        ebi::vcf::Version version;
+        try {
+            version = detect_version(line);
+        } catch (FileformatError * error) {
+            for (auto &output : outputs) {
+                output->write_error(*error);
+            }
+            return false;
+        }
         std::unique_ptr<Parser> validator = build_parser(sourceName, validationLevel, version, ploidy);
-        return is_valid_vcf_file(line, input, *validator, outputs);
+        return validate(line, input, *validator, outputs);
     }
 
-    Version extract_version(std::vector<char> line)
+    Version detect_version(const std::vector<char> &line)
     {
-        std::string common_substring{"##fileformat=VCFv4."};
-        std::vector<char>::iterator lineIt;
+        std::string common_substring{"##fileformat="};
+        std::vector<char>::const_iterator lineIt;
         std::string::const_iterator commonIt;
         std::tie(commonIt, lineIt) = std::mismatch(common_substring.begin(), common_substring.end(), line.begin());
         if (commonIt != common_substring.end()) {
-            // the fileformat is wrong, so take any version and let the parser report the error
-            return Version::v41;
+            throw new FileformatError{1, "The fileformat declaration is not valid (the line must start with "
+                    + common_substring + " and the value must be one of 'VCFv4.1', 'VCFv4.2' or 'VCFv4.3')"};
         }
 
-        if (*lineIt == '1') {
+        std::string provided{lineIt, line.end()};
+        util::remove_end_of_line(provided);
+
+        if (std::string{"VCFv4.1"}.compare(provided) == 0) {
             return Version::v41;
-        } else if (*lineIt == '2') {
+        } else if (std::string{"VCFv4.2"}.compare(provided) == 0) {
             return Version::v42;
-        } else if (*lineIt == '3') {
+        } else if (std::string{"VCFv4.3"}.compare(provided) == 0) {
             return Version::v43;
         } else {
-            // version not supported, so take any version and let the parser report the error
-            return Version::v41;
+            throw new FileformatError{1, "The fileformat declaration is not valid "
+                    "(the value must be one of 'VCFv4.1', 'VCFv4.2' or 'VCFv4.3')"};
         }
     }
 
-    bool is_valid_vcf_file(const std::vector<char> &firstLine,
+    bool validate(const std::vector<char> &firstLine,
                            std::istream &input,
                            ebi::vcf::Parser &validator,
                            std::vector<std::unique_ptr<ebi::vcf::ReportWriter>> &outputs)
@@ -176,6 +193,7 @@ namespace ebi
 
         return validator.is_valid();
     }
+
     void write_errors(const Parser &validator, const std::vector<std::unique_ptr<ReportWriter>> &outputs)
     {
         for (auto &error : validator.errors()) {
--- a/src/vcf/validator.cpp
+++ b/src/vcf/validator.cpp
@@ -20,10 +20,17 @@ namespace ebi
 {
   namespace vcf
   {
+    Version detect_version(const std::vector<char> &line);
+
+    std::unique_ptr<Parser> build_parser(std::string const &path,
+                                         ValidationLevel level,
+                                         Version version,
+                                         Ploidy ploidy);
+
     bool validate(const std::vector<char> &firstLine,
-                           std::istream &input,
-                           ebi::vcf::Parser &validator,
-                           std::vector<std::unique_ptr<ebi::vcf::ReportWriter>> &outputs);
+                  std::istream &input,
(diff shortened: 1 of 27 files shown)


src/vcf/validator.cpp
```diff
@@ -122,13 +120,52 @@ namespace ebi
...
+        std::vector<char> line;
+        ebi::util::readline(input, line);
+        ebi::vcf::Version version = extract_version(line);
+        std::unique_ptr<Parser> validator = build_parser(sourceName, validationLevel, version, ploidy);
+        return is_valid_vcf_file(line, input, *validator, outputs);
+    }
+
+    Version extract_version(std::vector<char> line)
+    {
+        std::string common_substring{"##fileformat=VCFv4."};
```
cyenyxe: Leaving the whole value (VCFv4.x) of the common substring would make this a bit more flexible. Some people are designing extensions to VCF [for proteomics](https://github.com/samtools/hts-specs/issues/160) and the fileformat declaration could change.

src/vcf/vcf_v41.ragel
```diff
@@ -226,7 +226,8 @@
     # Fileformat line
     action fileformat_error {
         ErrorPolicy::handle_error(*this,
-            new FileformatError{n_lines, "The fileformat declaration is not 'fileformat=VCFv4.1'"});
+            new FileformatError{n_lines,
+                "The fileformat declaration is not valid (value must be one of 'VCFv4.1', 'VCFv4.2' or 'VCFv4.3')"});
```
cyenyxe: I thought this message was going to be reported in `extract_version`. This is misleading because it states non-valid fileformats are valid for this state machine.

src/vcf/validator.cpp
```diff
@@ -122,13 +120,52 @@ namespace ebi
...
+        std::vector<char>::iterator lineIt;
+        std::string::const_iterator commonIt;
+        std::tie(commonIt, lineIt) = std::mismatch(common_substring.begin(), common_substring.end(), line.begin());
+        if (commonIt != common_substring.end()) {
+            // the fileformat is wrong, so take any version and let the parser report the error
```
cyenyxe: An error should be reported as soon as it happens. Letting the application run only increases chances for wrong behavior.

src/vcf/validator.cpp
```diff
@@ -122,13 +120,52 @@ namespace ebi
...
                            std::vector<std::unique_ptr<ebi::vcf::ReportWriter>> &outputs)
     {
+        std::vector<char> line;
+        ebi::util::readline(input, line);
+        ebi::vcf::Version version = extract_version(line);
+        std::unique_ptr<Parser> validator = build_parser(sourceName, validationLevel, version, ploidy);
+        return is_valid_vcf_file(line, input, *validator, outputs);
+    }
+
+    Version extract_version(std::vector<char> line)
```
cyenyxe: Maybe `detect_version` instead of `extract`? It is not extracted/removed from the source.

src/vcf/validator.cpp
```diff
@@ -20,6 +20,12 @@ namespace ebi
 {
   namespace vcf
   {
+    bool validate(const std::vector<char> &firstLine,
```
cyenyxe: If `build_parser`, `is_valid_vcf_file` and `detect_version` are declared in the .hpp file, shouldn't these other two methods also be for consistency?

src/vcf/validator.cpp
```diff
@@ -122,13 +126,63 @@ namespace ebi
...
+            for (auto &output : outputs) {
+                output->write_error(*error);
+            }
+            return false;
+        }
+        std::unique_ptr<Parser> validator = build_parser(sourceName, validationLevel, version, ploidy);
+        return validate(line, input, *validator, outputs);
+    }
 
+    Version detect_version(const std::vector<char> &line)
```
cyenyxe: What about something like this, using [std::basic_string::substr](http://en.cppreference.com/w/cpp/string/basic_string/substr)? Not sure if the indexes are right, please check.

```c++
std::string str(line.begin(), line.end());

if (str.substr(0, 13) == "##fileformat=") {
    if (str.substr(14) == "VCFv4.1") {
        return Version::v41;
    } else if (str.substr(14) == "VCFv4.2") {
        return Version::v42;
    } else if (str.substr(14) == "VCFv4.3") {
        return Version::v43;
    }
}

throw new FileformatError{1, "The fileformat declaration is not valid (the line must start with "
    + common_substring + " and the value must be one of 'VCFv4.1', 'VCFv4.2' or 'VCFv4.3')"};
}
```

test/vcf/debugulator_integration_test.cpp
```diff
@@ -40,7 +38,7 @@ namespace ebi
       std::vector<std::unique_ptr<ebi::vcf::ReportWriter>> reports;
       reports.emplace_back(report);
 
-      return is_valid_vcf_file(file, *validator, reports);
+      return vcf::is_valid_vcf_file(file, path.generic_string(), ebi::vcf::ValidationLevel::warning, 2, reports);
```
cyenyxe: Could you make the call to the `Ploidy` constructor explicit? It is not clear what the number means right now.

Applies to other classes, although in `ploidy_test.cpp` is it explicit.

src/vcf/validator.cpp
jmmut: I put this two in the cpp to hide them from public access (we agreed that only is_valid_vcf_file should be provided). As build_parser and detect_version are not used from the tests, I could move those 2 to the cpp as well to be a bit more consistent. Full public hpp or private declaration?

src/vcf/validator.cpp
jmmut: several points here:
- yes, it's a good idea to get rid of the `compare`.
- yes, one throw makes it cleaner
- not sure about substr (it mixes positions and counts, in fact the parameter should be 13 in both usages, which is unintuitive), I tried to avoid on purpose that kind of index harcoding.

what about:

```c++
        std::string common_substring{"##fileformat="};
        std::vector<char>::const_iterator lineIt;
        std::string::const_iterator commonIt;
        std::tie(commonIt, lineIt) = std::mismatch(common_substring.begin(), common_substring.end(), line.begin());

        if (commonIt == common_substring.end()) {
            std::string provided_version{lineIt, line.end()};
            util::remove_end_of_line(provided_version);

            if (provided_version == "VCFv4.1") {
                return Version::v41;
            } else if (provided_version == "VCFv4.2") {
                return Version::v42;
            } else if (provided_version == "VCFv4.3") {
                return Version::v43;
            }
        }

        throw new FileformatError{1, "The fileformat declaration is not valid (the line must start with "
                    + common_substring + " and the value must be one of 'VCFv4.1', 'VCFv4.2' or 'VCFv4.3')"};
```

if you really really want to remove that `std::mismatch`, we could do:
```c++
        std::string common_substring{"##fileformat="};
        std::string line{vector_line.begin(), vector_line.end()};

        if (line.substr(0, common_substring.size()) == common_substring) {
            std::string provided_version{line.substr(common_substring.size())};
            util::remove_end_of_line(provided_version);

            if (provided_version == "VCFv4.1") {
                return Version::v41;
            } else if (provided_version == "VCFv4.2") {
                return Version::v42;
            } else if (provided_version == "VCFv4.3") {
                return Version::v43;
            }
        }

        throw new FileformatError{1, "The fileformat declaration is not valid (the line must start with "
                    + common_substring + " and the value must be one of 'VCFv4.1', 'VCFv4.2' or 'VCFv4.3')"};
```

cyenyxe: Maybe it's the Java programmer in me, but I understand the latter easier. I feel the iterators make the code more convoluted in this case...

src/vcf/validator.cpp
cyenyxe: To make a free function private, it could also be declared in the hpp as static, or in the unnamed namespace. In this case we would be fine with either of them, but "static" forces internal linkage so I would be slightly in favor of that one.

TL;DR link: http://stackoverflow.com/a/11032872

And for more detailed discussions:
http://stackoverflow.com/questions/16864573/is-there-any-diference-between-defining-static-function-inside-anonymous-namespa
http://stackoverflow.com/questions/154469/unnamed-anonymous-namespaces-vs-static-functions
http://stackoverflow.com/questions/8460327/why-are-anonymous-namespaces-not-a-sufficient-replacement-for-namespace-static

jmmut: I'm ok with the static option, but I still see easier to understand that something has internal linkage having the declaration in the cpp. 

I see the need of this `static` usage when you only have a hpp file and don't have a cpp file, but here we do have both.

In general, facing two equivalent options I'd prefer the simpler one that doesn't confuse even the c++'s standard commitee. Even more with static, that can mean lots of different things depending on where it is.

cyenyxe: Fair enough, I can be fine with the forward declaration as those will never be part of the public API.

PR #66, opened 2017-04-15
Extracted record.cpp check blocks into private methods

Anishka0107:
In this PR, the various check blocks of the methods in record.cpp have been extracted into new private methods. 
Though it is unrelated to this PR, I have added the version checking for duplicate ID, because in earlier commits, duplicate IDs threw an error for v4.1 and v4.2, which was wrong. So, now it only throws error for v4.3 in case of duplicate IDs.

### DIFF
--- a/inc/vcf/file_structure.hpp
+++ b/inc/vcf/file_structure.hpp
@@ -171,19 +171,42 @@ namespace ebi
          * @throw std::invalid_argument
          */
         void check_chromosome() const;
-        
-        void check_chromosome_colons() const;
-        void check_chromosome_whitespaces() const;
 
         /**
-         * Checks that IDs are alphanumeric
+         * Checks that chromosome does not contain any colons
+         * 
+         * @throw std::invalid_argument
+         */
+        void check_chromosome_no_colons() const;
+
+        /**
+         * Checks that chromosome does not contain any white-spaces
+         * 
+         * @throw std::invalid_argument        
+         */
+        void check_chromosome_no_whitespaces() const;
+
+        /**
+         * Checks that IDs are alphanumeric and do not contain duplicate values
          * 
          * @throw std::invalid_argument
          */
         void check_ids() const;
         
-        void check_ids_semicolons_whitespaces() const;
-        void check_ids_duplicates() const;
+        /**
+         * Checks that ID contains no semicolons or white-spaces
+         * 
+         * @throw std::invalid_argument
+         */
+        void check_ids_no_semicolons_whitespaces() const;
+
+        /**
+         * Checks that ID contains no duplicate values in the same line
+         * 
+         * @throw std::invalid_argument
+         */
+        void check_ids_no_duplicates() const;
+
         /**
          * Checks the structure of an alternate allele and its accordance to the meta section
          * 
@@ -201,7 +224,12 @@ namespace ebi
          */
         void check_alternate_allele_structure(std::string const & alternate, RecordType type) const;
         
-        void check_alternate_allele_beginning(std::string const & alternate) const;
+        /**
+         * Check that alternates of the form <SOME_ALT> begin with DEL, INS, DUP, INV or CNV
+         * 
+         * @throw std::invalid_argument
+         */
+        void check_alternate_allele_symbolic_prefix(std::string const & alternate) const;
 
         /**
          * Checks that alternates of the form <SOME_ALT_ID> are described in the meta section
@@ -234,15 +262,25 @@ namespace ebi
         void check_info() const;
         
         /**
-         * Checks that format starts with GT
+         * Checks that format starts with GT and has no duplicate fields
          * 
          * @throw std::invalid_argument
          */
         void check_format() const;
-        
+
+        /**
+         * Checks that GT is the fisrt field in the FORMAT column
+         * 
+         * @throw std::invalid_argument
+         */
         void check_format_GT() const;
 
-        void check_format_duplicates() const;
+        /**
+         * Checks that format has no duplicate fields
+         * 
+         * @throw std::invalid_argument
+         */
+        void check_format_no_duplicates() const;
 
         /**
          * Checks that the samples in the record:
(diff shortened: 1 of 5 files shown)


Anishka0107:
@cyenyxe @jmmut I have not included comments for the new methods introduced in `record.hpp` and will put in them in a new commit to this PR once it is made sure that I am doing the extraction correctly. 
Also, I have not done the extraction for the method `check_samples()` yet as it seemed a bit complicated and will do that also in the next commit if needed.

inc/vcf/file_structure.hpp
```diff
@@ -172,13 +172,18 @@ namespace ebi
          */
         void check_chromosome() const;
         
+        void check_chromosome_colons() const;
```
cyenyxe: Some methods could be renamed to make their purpose clearer.

For instance, it is hard to know if this `check_chromosome_colons` should pass when colons are in the chromosome or not. Something like `check_chromosome_contains_no_colons` could be more obvious.

src/vcf/record.cpp
```diff
@@ -184,6 +188,23 @@ namespace ebi
         
     }
     
+    void Record::check_alternate_allele_beginning(std::string const & alternate) const
```
cyenyxe: To use more concrete terminology, this method could be renamed to `check_alternate_allele_symbolic_prefix`.

"Symbolic alternate allele" is the name specifically used for alternate alleles of the form <SOME_ALT>.

cyenyxe:
These changes look good to me. The options I see for extracting code from `check_samples` are:

1. [Initial check](https://github.com/Anishka0107/vcf-validator/blob/d9c47853b50b72d482ba4fa16c775508c2f8e4e7/src/vcf/record.cpp#L282-L284)
2. [This block](https://github.com/Anishka0107/vcf-validator/blob/d9c47853b50b72d482ba4fa16c775508c2f8e4e7/src/vcf/record.cpp#L293-L312) is not a check but is still quite self-contained
3. [This block](https://github.com/Anishka0107/vcf-validator/blob/d9c47853b50b72d482ba4fa16c775508c2f8e4e7/src/vcf/record.cpp#L319-L323) checks the number of fields in a sample column
4. In fact, [this whole block](https://github.com/Anishka0107/vcf-validator/blob/d9c47853b50b72d482ba4fa16c775508c2f8e4e7/src/vcf/record.cpp#L316-L362) could be turned into `check_sample` (notice the singular)

@jmmut please confirm you would be fine with these changes too.

src/vcf/record.cpp
```diff
@@ -340,20 +369,30 @@ namespace ebi
             if (allele == ".") { continue; } // No need to check missing alleles
 
             // Discard non-integer numbers
-            if (std::find_if_not(allele.begin(), allele.end(), isdigit) != allele.end()) {
-                throw new SamplesFieldBodyError{line, "Allele index " + allele + " is not an integer number",
-                                                "GT", ploidy};
-            }
+            check_samples_alleles_int(allele, ploidy);
```
jmmut: I would favor clearer (longer) method names like "check_samples_allele_is_integer" and remove the comments like "discard non-integer numbers"

inc/vcf/file_structure.hpp
```diff
@@ -171,15 +171,43 @@ namespace ebi
          * @throw std::invalid_argument
          */
         void check_chromosome() const;
-        
+
         /**
-         * Checks that IDs are alphanumeric
+         * Checks that chromosome does not contain any colons
+         * 
+         * @throw std::invalid_argument
```
Anishka0107: So, should I remove this particular line `@throw std::invalid_argument` from the comments for all the methods? In that case, would it be okay if I also changed the comment style from multi line comments `/*   */` to single line comments `//   ` as there would only be one line for most of the comments sections.

inc/vcf/file_structure.hpp
```diff
@@ -168,88 +168,185 @@ namespace ebi
...
+         */
+        void check_alternate_allele_symbolic_prefix(std::string const & alternate) const;
+
+        /**
          * Checks that alternates of the form <SOME_ALT_ID> are described in the meta section
          * 
-         * @throw std::invalid_argument
+         * @throw AlternateAllelesBodyError
          */
         void check_alternate_allele_meta(std::string const & alt_id,
```
jmmut: This method is implemented in validate_optional_policy, and not implemented or used as record method, so I would remove this prototype

src/vcf/record.cpp
```diff
@@ -281,79 +320,97 @@ namespace ebi
...
+        check_sample_subfields_count(i, subfields);
+        
+        std::vector<std::string> alleles;
+        // If the first format field is not a GT, then no alleles need to be checked
+        if (format[0] == "GT") {
+            util::string_split(subfields[0], "|/", alleles);
+                
+            // The allele indexes must not be greater than the total number of alleles
+            check_samples_alleles(alleles);
+        }
```
jmmut: @cyenyxe, I'm thinking if it's worth having this `if` inside its own method, so that `check_sample` reads naturally that it does 3 types of checks:
- check_sample_subfields_count
- check_sample_gt_if_present
- check_sample_subfields_cardinality_type

cyenyxe: Yes, that would be a nice method body. Even if the condition itself is not moved to another method, at least the `util::string_split` call should be moved inside `check_samples_alleles`.

Anishka0107:
@jmmut @cyenyxe 
Should I add a commit with the new changes you suggested above?

cyenyxe:
Yes please.

src/vcf/record.cpp
```diff
@@ -382,27 +378,29 @@ namespace ebi
         }
     }
 
-    void Record::check_samples_alleles(std::vector<std::string> const & alleles) const
+    void Record::check_sample_alleles(std::vector<std::string> subfields) const
     {
+        std::vector<std::string> alleles;
```
Anishka0107: Moved this line because `alleles` was needed for this particular method only and not anywhere else in `check_sample` method

inc/vcf/file_structure.hpp
```diff
@@ -168,88 +168,177 @@ namespace ebi
...
         
         /**
          * Check that the allele indexes in a sample are not greater than the total number of alleles
          * 
-         * @throw std::invalid_argument
+         * @throw SamplesFieldBodyError
          */
-        void check_samples_alleles(std::vector<std::string> const & alleles) const;
-        
+        void check_sample_alleles(std::vector<std::string> subfields) const;
```
jmmut: small slip here: use `const &` as the other methods to avoid copying

PR #68, opened 2017-05-02
Splitted fixer.hpp contents into fixer.cpp and fixer.hpp files

Anishka0107:
In this PR, the `inc/vcf/fixer.hpp` file which originally contained all the methods along with their definitions, has now been split into `inc/vcf/fixer.hpp` and `src/vcf/fixer.cpp` files.
The `fixer.hpp` file now contains the `Fixer` class along with the method declarations in it and the definitions are present in the `fixer.cpp` file.

### DIFF
--- a/inc/vcf/fixer.hpp
+++ b/inc/vcf/fixer.hpp
@@ -33,6 +33,8 @@ namespace ebi
       public:
         Fixer(std::ostream &output)
                 : line_number{0}, line{nullptr}, output(output), ignored_errors{0} { }
+        
+        virtual ~Fixer() { }
 
         void fix(size_t line_number, std::vector<char> &line, Error &error);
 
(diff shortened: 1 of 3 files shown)


inc/vcf/fixer.hpp
```diff
@@ -34,125 +34,35 @@ namespace ebi
...
-        virtual void visit(FilterBodyError &error) override
-        {
-            util::writeline(output, *line);
-            ignored_errors++;
-        }
+        virtual void visit(Error &error) override;
+        virtual void visit(MetaSectionError &error) override;
+        virtual void visit(HeaderSectionError &error) override;
+        virtual void visit(BodySectionError &error) override;
+        virtual void visit(NoMetaDefinitionError &error) override;
```
Anishka0107: I have added the `override` keyword here considering the fact that it is needed here & it might have got missed as a typo previously. If wrong please tell me, I will add another commit in that case.

jmmut: Yes, good spot! it's not strictly needed, but it is recommended because it won't allow compilation in the case that it is actually not overriding.

by the way, if we allow further inheriting (we do because we are using `virtual`), we should add a virtual destructor (`virtual ~Fixer() {}`) to avoid memory leaks due to polymorphic use http://stackoverflow.com/questions/461203/when-to-use-virtual-destructors.

PR #69, opened 2017-05-14
Adding meta line checks for predefined tags of INFO

Anishka0107:
Added the check for AA only. Will be adding more checks for the other tags when this is made perfect.


src/vcf/meta_entry.cpp
```diff
@@ -214,6 +214,13 @@ namespace ebi
             value["Type"] != "String") {
             throw new MetaSectionError{entry.line, "INFO metadata Type is not a Integer, Float, Flag, Character or String"};
         }
+
+        // Check for predefined tag AA
+        if (value["ID"] == "AA") {
+            if (value["Type"] != "String") {
```
Anishka0107: Separated this out from the upper if block instead of using `&&` operator. I think its better because we might need to add 2 or more checks for other tags and hence the separation avoids us from making the outer if block too length or complicated (using a lot of logical operators).

src/vcf/meta_entry.cpp
```diff
@@ -214,6 +214,13 @@ namespace ebi
...
+                throw new MetaSectionError{entry.line, "INFO with ID AA metadata Type is not a String"};
```
Anishka0107: Wasn't sure about the error message. Should it be `INFO AA metadata Type is not String`?

test/vcf/metaentry_test.cpp
```diff
@@ -570,6 +570,15 @@ namespace ebi
...
+
+        SECTION("INFO predefined tags") {
+            CHECK_THROWS_AS( (vcf::MetaEntry {
+                                1,
+                                "INFO",
+                                { {"ID", "AA"}, {"Number", "1"}, {"Type", "Integer"}, {"Description", "Genotype"} }
+                            }),
+                            vcf::MetaSectionError* );
+        }
     }
```
Anishka0107: Should a `CHECK_NOTHROWS` be added as well for `String` Type of AA?

src/vcf/meta_entry.cpp
cyenyxe: Yes, I think the one in the comment sounds better. It is hard to write error messages for these things 😅

src/vcf/meta_entry.cpp
cyenyxe: This block could be made more generic by using a `map<string, pair<string, string>>` which could be declared at namespace level. The keys would be IDss, and the values would be pairs with the Type and the message to display in case of error.

test/vcf/metaentry_test.cpp
cyenyxe: Yes please, it is a good practice to test both the positive and negative cases.

src/vcf/meta_entry.cpp
```diff
@@ -214,6 +214,13 @@ namespace ebi
             value["Type"] != "String") {
             throw new MetaSectionError{entry.line, "INFO metadata Type is not a Integer, Float, Flag, Character or String"};
         }
+
+        // Check for predefined tag AA
+        if (value["ID"] == "AA") {
```
jmmut: Now you are touching this method, I think it would be good to use temporal variables for `value` as const references like this: `const std::string &type = value["Type"];`. That would improve speed a bit and decrease risk of misspelling "Type" and the other ones. Maybe we could even extract those hardcoded strings?

src/vcf/meta_entry.cpp
jmmut: I like the `map` idea. how should it be declared and defined? I always forget the recommended practice for namespace constants. const static? and definition with curly braces?

src/vcf/meta_entry.cpp
cyenyxe: I agree with the idea but would it leave it for another PR. Depending on the libraries we use to extract them we could even leave open the door for easy localization.

src/vcf/meta_entry.cpp
Anishka0107: Oh this idea is great, but what about Number? I guess that would be required for a few other tags, shouldn't I include it?

src/vcf/meta_entry.cpp
Anishka0107: Should I include this change in the next commit😅?

Anishka0107: I haven't included this thing as I thought it would have to be changed at various places. If needed I can do this in another PR?

src/vcf/meta_entry.cpp
```diff
@@ -214,6 +220,22 @@ namespace ebi
             value["Type"] != "String") {
             throw new MetaSectionError{entry.line, "INFO metadata Type is not a Integer, Float, Flag, Character or String"};
         }
+
+        for (const auto & t : info_type) {
+            if (value["ID"] == t.first) {
+                if (value["Type"] != t.second.first) {
```
jmmut: reuse the `number_field` variable (from line 206) for this method, and do the same for `value["Type"]`.

Anishka0107:
If the changes till now are fine, can I start adding new checks as well?

test/input_files/v4.1/passed/complexfile_passed_000.vcf
```diff
@@ -51,9 +51,9 @@
...
...:0.00,-0.90,-18.30	0|0:0.000:0,0,0	0|0:0.000:0.00,-3.00,-60.80	0|0:0.000:0,0,0	0|0:0.000:0.00,-0.30,-6.10	0|0:0.000:0,0,0	0|0:0.000:0.00,-0.60,-12.20	0|0:0.000:0.00,-0.60,-12.20	0|0:0.000:0.00,-0.90,-18.30	0|0:0.050:0,0,0	0|0:0.000:0.00,-0.90,-18.30	0|0:0.000:0.00,-0.90,-18.30	0|0:0.000:0,0,0	0|0:0.000:0.00,-1.50,-30.50	0|0:0.000:0.00,-0.30,-6.10	0|0:0.000:0,0,0	0|0:0.000:0.00,-0.30,-6.10	0|0:0.000:0,0,0	0|0:0.000:0,0,0	0|0:0.000:0.00,-1.20,-24.40	0|0:0.000:0,0,0	0|0:0.000:0,0,0	0|0:0.000:0.00,-0.30,-6.10	0|0:0.000:0,0,0	0|0:0.000:0.00,-0.60,-12.20	0|0:0.000:0,0,0	0|0:0.000:0,0,0	0|0:0.000:0,0,0	0|0:0.000:0,0,0	0|0:0.050:0,0,0	0|0:0.000:0,0,0	0|0:0.000:0,0,0	0|0:0.000:0.00,-1.20,-24.40	0|0:0.000:0.00,-0.30,-6.10	0|0:0.000:0,0,0	0|0:0.000:0.00,-0.30,-6.10	0|0:0.050:0,0,0	0|0:0.000:0.00,-0.90,-18.30	0|0:0.000:0.00,-0.60,-12.20	0|0:0.000:0.00,-0.30,-6.10	0|0:0.050:0,0,0	0|0:0.000:0.00,-0.60,-12.20	0|0:0.000:0.00,-1.20,-24.40	0|0:0.000:0.00,-0.60,-12.20	0|0:0.000:0.00,-0.30,-6.10	0|0:0.000:0,0,0	0|0:0.000:0,0,0	0|0:0.000:0,0,0	0|0:0.000:0.00,-0.30,-6.10	0|0:0.050:0,0,0	0|0:0.000:0,0,0	0|0:0.000:0.00,-0.60,-12.20	0|0:0.000:0,0,0	0|0:0.000:0.00,-0.30,-6.10	0|0:0.000:0,0,0	0|0:0.050:0,0,0
```
Anishka0107: The change I made here and in any other line like this one is just adding 2 values for the data line with INFO AC.
I changed `AC=10` to `AC=10,11`

src/vcf/meta_entry.cpp
```diff
@@ -216,16 +219,28 @@ namespace ebi
...
+        else if (entry.source->version == Version::v43) {
+            version_info_type_map = info_type_v43;
+            version_info_number_map = info_number_v43;
+        }
+        
+        for (const auto & t : version_info_type_map) {
             if (value["ID"] == t.first) {
                 if (type_field != t.second.first) {
-                    throw new MetaSectionError{entry.line, t.second.second};
+                        throw new MetaSectionError{entry.line, t.second.second};
```
jmmut: previous indentation was correct

src/vcf/meta_entry.cpp
```diff
@@ -216,16 +219,28 @@ namespace ebi
...
                 }
             }
         }
 
-        for (const auto & n : info_number) {
+        for (const auto & n : version_info_number_map) {
```
jmmut: when this part is extracted to another function, change variable names to `number` and `info_numbers`. (and `type` and `info_types` in the previous `for`)

src/vcf/meta_entry.cpp
```diff
@@ -216,16 +219,28 @@ namespace ebi
...
 
-        for (const auto & t : info_type) {
+        if (entry.source->version == Version::v41 || entry.source->version == Version::v42) {
+            version_info_type_map = info_type_v41_v42;
+            version_info_number_map = info_number_v41_v42;
+        }
+        else if (entry.source->version == Version::v43) {
+            version_info_type_map = info_type_v43;
+            version_info_number_map = info_number_v43;
+        }
```
jmmut: I don't like much that map copy. In this case I think it's better to extract the two `for`s to a function, and here do the call with the maps of one version or the other:
```
if (entry.source->version == Version::v41 || entry.source->version == Version::v42) {
    check_predefined_type_and_number(info_type_v41_v42, info_number_v41_v42);
} else {
    check_predefined_tags_type_and_number(info_type_v43, info_number_v43);
}
```

test/vcf/metaentry_test.cpp
```diff
@@ -38,12 +38,12 @@ namespace ebi
             
             SECTION ("It should work with any ID and source")
             {
-                CHECK_NOTHROW( (vcf::MetaEntry { 1, "reference" }) );
+                CHECK_NOTHROW( (vcf::MetaEntry { 1, "reference", std::make_shared<vcf::Source>(source) }) );
```
jmmut: I'd say the constructor call will be easier if you replace the type of the source to `shared_ptr`:
```
std::shared_ptr<Source> source{
        std::make_shared<vcf::Source>(
                "Example VCF source",
                vcf::InputFormat::VCF_FILE_VCF | vcf::InputFormat::VCF_FILE_BGZIP,
                vcf::Version::v41,
                vcf::Ploidy{2},
                {},
                { "Sample1", "Sample2", "Sample3" })};
```
and then I would do a find&replace of `std::make_shared<vcf::Source>(source)` to `source`. the same change applies to record_test.cpp.

also, the indentation of the `SECTION`s in the 2 first `TEST_CASE`s is wrong.

Issue #19, opened 2016-04-29
INFO and SAMPLE floating point fields range is not checked


cyenyxe:
Some integer number fields are checked to guarantee their zero or positive, but this doesn't happen for floating-point fields. Also, those fields that represent frequencies are not checked to be between 0 and 1.

This fix will involve un-commenting the test file `failed_body_info_004.vcf`.

jmmut:
There are some fields like GL that are logarithms of frequencies, so its range is (-inf, 0], and this range is not described in the meta entry of the field, so we can only do range checks for the reserved fields described in the spec.

jmmut:
done partially in #72. except for some fields that are not unambiguously specified (https://github.com/samtools/hts-specs/issues/224 and the ones about likelihood such as PL or GL)

Anishka0107:
Done partly in https://github.com/EBIvariation/vcf-validator/pull/89

jmmut:
another bug: FORMAT GP in 4.2 is not a probability [0, 1] as in 4.3, it's a phred-scale (https://genome.sph.umich.edu/wiki/Phred_scale) which I think it's [0, +inf)

tcezard:
Tracked in EVA-2051: closing

tcezard:
tracked in EVA-2052: closing

PR #72, opened 2017-06-10
Extracted semantic checks from ragel and added strict validation

Anishka0107:


src/vcf/record.cpp
```diff
@@ -241,6 +243,45 @@ namespace ebi
...
+                        check_predefined_type("INFO", field.first, value, info_v41_v42);
+                    check_predefined_number("INFO", field.first, values.size(), info_v41_v42);
+                } else {
+                    for (auto & value : values)
+                        check_predefined_type("INFO", field.first, value, info_v43);
+                    check_predefined_number("INFO", field.first, values.size(), info_v43);
+                }
+            }
+
+            // strict validation for tags
```
cyenyxe: Extract these checks to another method.

src/vcf/record.cpp
```diff
@@ -241,6 +243,45 @@ namespace ebi
...
+            if (!found) {
+                std::vector<std::string> values;
+                util::string_split(field.second, ",", values);
+                if (source->version == Version::v41 || source->version == Version::v42) {
+                    for (auto & value : values)
+                        check_predefined_type("INFO", field.first, value, info_v41_v42);
+                    check_predefined_number("INFO", field.first, values.size(), info_v41_v42);
+                } else {
+                    for (auto & value : values)
+                        check_predefined_type("INFO", field.first, value, info_v43);
```
cyenyxe: Wouldn't it be possible to use the existing `check_field_type` and `check_field_cardinality` methods? The expected type or number would need to be extracted from the map before the method call, but otherwise I think they should be fine. That would also bring more consistency to the error reports and prevent code to be changed, for instance, in `check_field_type` but not in `check_predefined_type`.

inc/vcf/file_structure.hpp
```diff
@@ -67,160 +67,82 @@ namespace ebi
...
+            { "CIGAR", { "String", "A" } },
+            { "DB", { "Flag", "0" } },
+            { "DP", { "Integer", "1" } },
+            { "END", { "Integer", "1" } },
+            { "H2", { "Flag", "0" } },
+            { "H3", { "Flag", "0" } },
+            { "MQ", { ".", "1" } },
+            { "MQ0", { "Integer", "1" } },
+            { "NS", { "Integer", "1" } },
+            { "SB", { ".", "." } },
```
Anishka0107: Okay! But, I'm a bit doubtful here.
Could you please take a look at https://github.com/EBIvariation/vcf-validator/blob/develop/src/vcf/vcf_v43.ragel#L255. In the original ragel checks we are using the `any_number` regex, which means it would match any floating point or integer number but not a String. And the error we throw is : https://github.com/EBIvariation/vcf-validator/blob/develop/src/vcf/vcf.ragel#L565-L571
The same goes for `MQ` in https://github.com/EBIvariation/vcf-validator/blob/develop/src/vcf/vcf_v43.ragel#L252.

Also, if the " . " related checks are removed, we would have to delete these test files from each version : 
[SB related](https://github.com/EBIvariation/vcf-validator/blob/develop/test/input_files/v4.1/failed/failed_body_info_026.vcf)
[MQ related](https://github.com/EBIvariation/vcf-validator/blob/develop/test/input_files/v4.1/failed/failed_body_info_021.vcf)

So, what should be done - removing the " . " related check and the associated test files OR keeping it?

cyenyxe: We can't say for sure that SB must be a number, because it is not defined in the specification. So yes, those files should be removed...

On the other hand, check the file referenced in #19, because it could be possible to uncomment it after the last changes you have made. Ignore the comment about checks in the Ragel code because, as you can appreciate, we end up deciding not to do it that way :)

src/vcf/record.cpp
```diff
@@ -512,9 +595,20 @@ namespace ebi
                     }
                 } else if (type == "String") {
                     // ...do nothing, it is guaranteed it will be a string
-                } 
+                }
             } catch (...) {
-                raise(std::make_shared<Error>(line, field + " does not match the meta specification Type=" + type));
+                if (status) {
```
jmmut: Ok, after thinking a while about this, I suggest the next refactor. The main idea is to remove the  `status` flag and make each function receive only the information it needs. Please consider the next 2 functions:
```    
    void check_value_type(const std::string &type, const std::string &value) {
        if (type == "Integer") {
            // ...try to cast to int
            std::stoi(value);
            // ...and also check it's not a float
            if (std::fmod(std::stof(value), 1) != 0) {
                throw std::invalid_argument("Float provided instead of Integer");
            }
        } else if (type == "Float") {
            // ...try to cast to float
            try {
                std::stof(value);
            } catch (std::out_of_range) {
                // It maybe a subnormal number
                std::stold(value);
            }
        } else if (type == "Flag") {
            int numeric_value = std::stoi(value);
            if (numeric_value != 0 && numeric_value != 1) {
                throw std::invalid_argument("A flag value must be 0 or 1");
            }
            // If no flag is provided then there is nothing to check
        } else if (type == "Character") {
            // ...check the length is 1
            if (value.size() > 1) {
                throw std::invalid_argument("There can be only one character");
            }
        } else if (type == "String") {
            // ...do nothing, it is guaranteed it will be a string
        }
    }

    void Record::check_field_type(std::string const & field,
                                  std::vector<std::string> const & values,
                                  std::string const & type) const
    {
        // To check the field type...
        for (auto & value : values) {
            if (value == ".") { continue; }
            
            try {
                check_value_type(type, value);
            } catch (std::exception &typeError) {
                throw std::invalid_argument(field + " does not match the meta specification Type=" + type + ". " 
                                                    + typeError.what());
            }
        }
    }
```
As you see, the first one only knows about one value (if a field is "AC=3,4,5" this function will receive "Integer" and "3", for instance), and consequently, it raises an exception containing that information. In the second function, we are in the context of a type check for several values, so it add that information to the exception message and rethrows. finally, in some place like `check_info`, the second function will be called, and will catch an exception and will add to the message that it is a "INFO" check (that's the only place that should know it), and only then build an Error from our hierarchy, a `InfoBodyError` in this case, reusing the message from the inner exception.

A similar thing can be done for `check_predefined_tag`.

src/vcf/record.cpp
```diff
@@ -518,97 +537,70 @@ namespace ebi
...
-                    }
-                    if (tag_field == "INFO") {
-                        throw new InfoBodyError{line, message, field};
-                    } else {
-                        throw new FormatBodyError{line, message};
-                    }
-                }
+                check_value_type(type, value, message);
+            } catch (std::exception &typeError) {
+                raise(std::make_shared<Error>(line, field + " does not match the meta specification Type=" + type + ". " + message));
```
jmmut: Ok, I think we can do a final refactor here. the purpose is to remove that message returned by reference to build the exception. I was thinking in putting a bit more detail in the inner functions like `check_value_type` and `check_field_cardinality`.

This is a good opportunity to address a suggestion a user made some time ago, that the messages would be more useful if they had (on top of the explanation of the error) the actual offending values (like "AC=2,3,4").

To guide the refactor, try to remove the `try-catch` in this function (and maybe `check_predefined_tag` if it's possible?), so at the end we have these `try`s and `catch`es:
- in `check_field_cardinality` and `check_value_type` throw std::invalid_argument with a message that include the explanation (like "Float provided instead of Integer") but also the exact offending value (`value`) and stating clearly  that it is a problem regarding the type or cardinality. Something like `value + " (float) was provided as value of Type=" + type` 
- in check_info,` catch(const std::invalid_argument &ex)`: throw an InfoBodyError with something like "INFO " + field.first + "=" + field.second + ": " + ex.what();
- in check_info, but for the predefined tags,` catch(const std::invalid_argument &ex)`: throw InfoBodyError with "INFO (defined in the specification) " + field.first + "=" + field.second + ": " + ex.what();
- in check_sample_subfields_cardinality_type, throw SamplesFieldBodyError similar message including the sample number

Anishka0107: I haven't changed the `std::shared_ptr<Error>` to `const std::invalid_argument &` yet! Will be doing it in another commit/PR when this part gets perfected.

src/vcf/record.cpp
```diff
@@ -274,9 +289,54 @@ namespace ebi
...
+                                      std::map<std::string, std::pair<std::string, std::string>> const & tags) const
+    {
+        auto iterator = tags.find(field_key);
+        if (iterator != tags.end()) {
+            try {
+                check_field_cardinality(field_key, values, iterator->second.second);
+                check_field_type(values, iterator->second.first);
+            } catch (std::shared_ptr<Error> ex) {
+                raise(std::make_shared<Error>(line, field_key + "=" + field_value
+                                + " is not compliant with the VCF specification" + ex->message));
```
Anishka0107: The error message thrown would be something like `INFO AC=2,3,4,5 is not complaint with the VCF specification Number=G. Contains 4 values, expected 2` 
I'm not sure about the message, especially this `Number=X` or `Type=Y` part, but I added it to avoid some problem in the code and also because adding this would prompt the user to check that the predefined tag does not contain a meta definition and it would be good to add it. Or say, they will get to know that the Number has been defined as A, so that's why the expected number of values is 2 (the number of alt alleles present in the line).

src/vcf/record.cpp
```diff
@@ -459,62 +519,73 @@ namespace ebi
...
         if (!number_matches) {
-            raise(std::make_shared<Error>(line, field + " does not match the meta specification Number=" + number +
-                    ", expected " + std::to_string(expected) + " values"));
+            raise(std::make_shared<Error>(line, " specification Number=" + number + ". Contains " + std::to_string(values.size()) + " values, expected " + std::to_string(expected)));
         }
     }
 
-    void Record::check_field_type(std::string const & field,
-                                  std::vector<std::string> const & values,
+    void Record::check_value_type(std::string const & type, std::string const & value, std::string & message) const {
```
cyenyxe: I still find the format of these error messages a bit inconsistent. This is the output for a synthetic VCF file:

```
Line 7: INFO INT=1.5 does not match the meta specification Type=Integer. 1.5 (float) provided instead of Integer
Line 8: INFO VALIDATED=2 does not match the meta specification Type=Flag. A flag value must be 0 and 1
Line 9: INFO FLOAT=lalala does not match the meta specification Type=Float.
```

For the first error, the second sentence could be `An integer must not contain decimal digits`. 
For the second error, a flag value must be "0, 1 or none".
For all errors, the last sentence should end with a full stop.

src/vcf/record.cpp
```diff
@@ -459,62 +519,73 @@ namespace ebi
...
+            }
+        }
+    }
+
+    void Record::check_field_integer_range(std::string const & field, std::vector<std::string> const & values) const {
+        for (auto & value : values) {
+            if (value == ".") { continue; }
+
+            if (std::stoi(value) < 0) {
+                raise(std::make_shared<Error>(line, field + " value is negative. It must be a non-negative integer"));
```
cyenyxe: `field + " value must be a non-negative integer number"` for simplicity?

PR #73, opened 2017-06-22
Meta header checks for INFO and FORMAT SV tags

Anishka0107:
This PR aims to add meta header checks for all Info and Format SV tags. The Type and Number of the meta header definition would be checked for SV specific tags.


Anishka0107:
The remaining tests which could be added are the negative CHECK assertions, passed test file(s), failed test file(s).

inc/vcf/file_structure.hpp
```diff
@@ -85,7 +85,31 @@ namespace ebi
             // TODO : SB metadata Type and Number is .
             { "SOMATIC", { "Flag", "0" } },
             { "VALIDATED", { "Flag", "0" } },
-            { "1000G", { "Flag", "0" } }
+            { "1000G", { "Flag", "0" } },
+            // tags specific to SVs are listed below
```
cyenyxe: Please sort these alphabetically to facilitate searching. Applies to the other maps too.

PR #76, opened 2017-06-30
Implement duplicates check for other fields

Anishka0107:


src/vcf/record.cpp
```diff
@@ -472,6 +470,19 @@ namespace ebi
...
     }
 
+    void Record::check_no_duplicates(std::vector<std::string> const & values) const
+    {
+        if (values.size() > 1) {
+            std::map<std::string, int> counter;
+            for (auto & value : values) {
+                counter[value]++;
+                if (counter[value] >= 2) {
+                    raise(std::make_shared<Error>(line, " must not have duplicate "));
```
cyenyxe: The way this message needs to be used from other methods is a bit ugly. What about just throwing an `std::invalid_argument`, like in `check_value_type`, then build the whole message in the other methods?

src/vcf/record.cpp
```diff
@@ -259,6 +273,17 @@ namespace ebi
        }
     }
     
+    void Record::check_info_no_duplicates() const
+    {
+        if (source->version == Version::v43 && info.size() > 1) {
+            for (auto & in : info) {
+                if (info.count(in.first) > 1) {
+                    throw new InfoBodyError{line, "INFO must not have duplicate keys"};
```
Anishka0107: Another way we could do it is using the `Record::check_no_duplicates()` method. But for that we'd have to use a vector for the keys and then construct the counting map. So I think using this is better instead.

Anishka0107: Another way we could do it is using the `Record::check_no_duplicates()` method. But for that we'd have to use a vector for the keys and then construct the counting map. So I think using this is better instead.

inc/vcf/file_structure.hpp
```diff
@@ -380,21 +380,42 @@ namespace ebi
...
+        /**
+         * Check that FILTER values are non-zero, as 0 is reserved and must not be used
+         *
+         * @throw FilterBodyError
+         */
+        void check_filter_not_zero() const;
+
+        /**
+         * Checks that all the INFO fields are listed in the meta section, their number and 
+         * type match those specifications, and conatin no duplicates
```
cyenyxe: Typo in "contain"

inc/vcf/file_structure.hpp
```diff
@@ -496,6 +517,13 @@ namespace ebi
         void check_sample_alleles_range(std::string const & allele, long ploidy) const;
 
         /**
+         * Checks that fields have no duplicates
```
cyenyxe: The function is so generic that it doesn't refer to a field at all, so maybe documentation like this?

`Checks that a list contains no duplicates`

src/vcf/record.cpp
```diff
@@ -472,6 +505,19 @@ namespace ebi
...
     }
 
+    void Record::check_no_duplicates(std::vector<std::string> const & values) const
+    {
+        if (values.size() > 1) {
+            std::map<std::string, int> counter;
+            for (auto & value : values) {
+                counter[value]++;
+                if (counter[value] >= 2) {
+                    throw std::invalid_argument(" must not have duplicate ");
```
cyenyxe: This message is unclear just by itself. I think it would be best to throw the exception without any message, then build the whole message in the methods outside. Applies to ID, FILTER and FORMAT that I can see.

test/vcf/record_test.cpp
```diff
@@ -510,5 +510,58 @@ namespace ebi
...
+                                "A",
+                                { "T", "C" },
+                                1.0, 
+                                { "PASS" }, 
+                                { {"AN", "12"}, {"AN", "15"} }, 
+                                { "DP" }, 
+                                { "12" },
+                                source}),
+                            vcf::InfoBodyError*);
+        }
```
cyenyxe: Could `SECTION("Duplicate FORMATs")` be placed after INFO to follow the same order as the VCF format? It could make a bit easier to find where the tests are.

PR #74, opened 2017-06-25
Add functionality to check that predefined tags in data lines match the VCF spec

Anishka0107:
Added check to ensure predefined FORMAT tags in data lines which haven't been defined in the meta header specification, match the Type and Number specified in the VCF spec.
Would go through each subfield in each sample field and throw error in a manner similar to INFO predefined tags for inconsistencies.
Also fixed breaking tests. A few tests which were originally passing, threw errors because the predefined FORMAT tags didn't match the spec, so changed those values to ensure these files pass the test.
Added new strict validation checks for FORMAT tags.
Added new strict validation checks for SV related tags.

PR #77, opened 2017-07-12
Extract string literals to constants, as they are prone to typos

Anishka0107:


Anishka0107:
I have to do the same for a few other files (basically the code & new files added in PR #74).

src/vcf/record.cpp
```diff
@@ -344,33 +344,33 @@ namespace ebi
...
+            auto it = info.find(IMPRECISE);
             if (it != info.end() && it->second == "0") {
                 auto expected = std::to_string(position + reference_allele.length() - 1);
                 if (field_value != expected) {
                     throw new InfoBodyError{line, "INFO END=" + field_value + " value must be equal to \"POS + length of REF - 1\" for a precise variant (where IMPRECISE is set to 0), expected " + expected, field_key};
                 }
             }
-        } else if (field_key == "SVLEN" && values.size() == alternate_alleles.size()) {
+        } else if (field_key == SVLEN && values.size() == alternate_alleles.size()) {
             bool is_ref_symbolic = true;
```
jmmut: this is used sometimes properly and sometimes as `is_ref_not_symbolic`. later in this function, the code under `if (is_ref_symbolic && is_alt_symbolic[i])` is in fact what we should do if the alleles are not symbolic, which is taking the difference in allele string sizes. also, the reference can not be a symbolic allele (the ragel code checks that under the record_ref machine)

src/vcf/record.cpp
```diff
@@ -344,33 +344,33 @@ namespace ebi
...
                 if (!boost::regex_match(value, cigar_string)) {
                     throw new InfoBodyError{line, "INFO CIGAR=" + field_value + " value is not an alphanumeric string compliant with the SAM specification", field_key};
                 }
             }
-        } else if (field_key == "END") {
-            auto it = info.find("IMPRECISE");
+        } else if (field_key == END) {
+            auto it = info.find(IMPRECISE);
             if (it != info.end() && it->second == "0") {
                 auto expected = std::to_string(position + reference_allele.length() - 1);
```
jmmut: I think I had an oversight here, shouldn't this be done only if reference is not symbolic?

test/vcf/predefined_format_tags_test.cpp
```diff
@@ -42,12 +42,12 @@ namespace ebi
                             "chr1",
                             123456,
                             { "id123" },
-                            "A",
+                            vcf::A,
```
jmmut: This usages (bases string) should not be extracted as the same constant for "cardinality A". revert all usages in this file, please

test/vcf/predefined_info_tags_test.cpp
```diff
@@ -42,12 +42,12 @@ namespace ebi
                             "chr1",
                             123456,
                             { "id123" },
-                            "A",
+                            vcf::A,
```
jmmut: same as predefined_format_tags_test, this usages (bases string) should not be extracted as the same constant for "cardinality A". revert all usages in this file, please

test/vcf/record_test.cpp
```diff
@@ -35,54 +35,54 @@ namespace ebi
...
                 1,
-                "INFO",
+                vcf::INFO,
                 {
-                    { "ID", "AF" },
-                    { "Number", "A" },
-                    { "Type", "Float" },
-                    { "Description", "Allele frequency" }
+                    { vcf::ID, vcf::AF },
+                    { vcf::NUMBER, vcf::A },
```
jmmut: except for this usage, which is ok, revert all the other usages of "A" in this file as well

test/vcf/record_test.cpp
```diff
@@ -95,12 +95,12 @@ namespace ebi
                                 "chr1", 
                                 123456, 
                                 { "id123", "id456" }, 
-                                "A", 
-                                { "AC", "AT" }, 
+                                vcf::A, 
+                                { vcf::AC, "AT" },
```
jmmut: revert back AC to hardcoded string in this file, as it is used as a bases string too.

src/vcf/record.cpp
jmmut: nevermind, remembered that references cannot be symbolic

src/vcf/meta_entry.cpp
```diff
@@ -83,172 +83,172 @@ namespace ebi
...
         if (iterator != tags.end()) {
             // Determine the required value of the key based on whether we are checking for Type or Number
-            std::string key_value = (key_field == "Type" ? iterator->second.first : iterator->second.second);
+            std::string key_value = (key_field == TYPE ? iterator->second.first : iterator->second.second);
             // If the required value is a "." (dot), do nothing
             // Or if the required value does not match the value provided in the vcf file, throw an error
-            if (key_value != "." && key_value != value[key_field]) {
-                std::string message = tag_field + " " + value["ID"] + " metadata " + key_field + " is not " + key_value;
+            if (key_value != MISSING_VALUE && key_value != value[key_field]) {
+                std::string message = tag_field + " " + value[ID] + " metadata " + key_field + " is not " + key_value;
```
Anishka0107: I agree they sound much better!

test/vcf/predefined_format_tags_test.cpp
Anishka0107: Oh sorry, I did a search & replace, so got this wrong

test/vcf/record_test.cpp
```diff
@@ -481,7 +481,7 @@ namespace ebi
                 vcf::INFO,
                 {
                     { vcf::ID, vcf::AF },
-                    { vcf::NUMBER, vcf::A },
+                    { vcf::NUMBER, "A" },
```
jmmut: oops, this usage was correct as vcf::A, I thought there was only one usage of this at the beginning of this file

test/vcf/record_test.cpp
```diff
@@ -110,11 +110,11 @@ namespace ebi
                                 123456, 
                                 { }, 
                                 "A", 
-                                { "AC", "AT" }, 
+                                { vcf::AC, "AT" },
```
jmmut: all the usages of vcf::AC in this file should be "AC" actually

src/vcf/record.cpp
```diff
@@ -416,6 +392,19 @@ namespace ebi
         }
     }
 
+    bool Record::check_alt_not_symbolic(size_t i) const
```
cyenyxe: Renaming `i` to `allele_index` would clarify its purpose.

inc/vcf/string_constants.hpp
```diff
@@ -0,0 +1,173 @@
...
+  namespace vcf
+  {
+
+    // VCF versions
+    const std::string VCF_V41 = "VCFv4.1";
+    const std::string VCF_V42 = "VCFv4.2";
+    const std::string VCF_V43 = "VCFv4.3";
+
+    // validator and debugulator strings
+    const char STDIN[] = "stdin";
```
cyenyxe: Does boost require the arguments to be `const char*` instead of `string`?

Anishka0107: Yes

PR #78, opened 2017-07-16
Add new data line fixing functionality in the debugulator

Anishka0107:
The fixes added are :

- #### Fix SVLEN value for non-symbolic ALTs
If the alternate allele in the variant is not symbolic, & SVLEN is integral but not equal to `len(ALT) - len(REF)`, set it to this value.

- #### Fix END value for precise variants
If a variant is flagged as precise (using IMPRECISE=0), & END is an integer nut not equal to `POS + len(REF) - 1`, set it to this value.

- #### Removal of duplicates in ID
The fix is to keep the first duplicated ID, and remove all consequent ones.

- #### Removal of duplicates or `0`'s in FILTER
The fix is to remove all 0's (if present) and keep the first duplicate filter, & remove the consequent ones.

- #### Removal of duplicates in INFO
The fix here is :
i) remove all the duplicate fields, if there are different values for the same key (we cannot give preference to any particular field)
ii) keep first and remove all consequent occurrences of a field if for the key, all the duplicate fields have the same value

- #### Removal of duplicates in FORMAT
The fix is: 
i) remove all duplicate fields from the format and sample(s) columns, if there are different values in one or more of the samples
ii) keep first and remove all consequent occurrences of a field from format and sample(s) if each sample has the same value for that field

### DIFF
--- a/inc/vcf/error.hpp
+++ b/inc/vcf/error.hpp
@@ -293,7 +293,7 @@ namespace ebi
                       const std::string &field = "",
                       const std::string &expected_value = "")
                 : BodySectionError{line, message}, error_fix{error_fix}, field{field}, expected_value{expected_value} {
-                    if (error_fix == ErrorFix::RECOVERABLE_VALUE && expected_value == "") {
+                    if (error_fix == ErrorFix::RECOVERABLE_VALUE && expected_value.empty()) {
                         throw std::invalid_argument{"An error with recoverable value must provide a non-empty expected value"};
                     }
                 }
@@ -307,11 +307,16 @@ namespace ebi
     #pragma db object
     struct FormatBodyError : public BodySectionError
     {
-        using BodySectionError::BodySectionError;
-        FormatBodyError() : FormatBodyError{0} {}
-        FormatBodyError(size_t line) : FormatBodyError{line, "Format is not a colon-separated list of alphanumeric strings"} { }
+        FormatBodyError(size_t line = 0,
+                        const std::string &message = "Format is not a colon-separated list of alphanumeric strings",
+                        ErrorFix error_fix = ErrorFix::IRRECOVERABLE_VALUE,
+                        const std::string &field = "")
+                : BodySectionError{line, message}, error_fix{error_fix}, field{field} { }
         virtual ~FormatBodyError() override { }
         virtual void apply_visitor(ErrorVisitor &visitor) override { visitor.visit(*this); }
+
+        ErrorFix error_fix;
+        const std::string field;
     };
     #pragma db object
     struct SamplesBodyError : public BodySectionError
--- a/inc/vcf/error.hpp
+++ b/inc/vcf/error.hpp
@@ -234,14 +234,12 @@ namespace ebi
     {
         IdBodyError(size_t line = 0,
                     const std::string &message = "ID is not a single dot or a list of strings without semicolons or whitespaces",
-                    ErrorFix error_fix = ErrorFix::IRRECOVERABLE_VALUE,
-                    const std::string &field = "")
-                : BodySectionError{line, message}, error_fix{error_fix}, field{field} { }
+                    ErrorFix error_fix = ErrorFix::IRRECOVERABLE_VALUE)
+                : BodySectionError{line, message}, error_fix{error_fix} { }
         virtual ~IdBodyError() override { }
         virtual void apply_visitor(ErrorVisitor &visitor) override { visitor.visit(*this); }
 
         ErrorFix error_fix;
-        const std::string field;
     };
     #pragma db object
     struct ReferenceAlleleBodyError : public BodySectionError
@@ -309,14 +307,12 @@ namespace ebi
     {
         FormatBodyError(size_t line = 0,
                         const std::string &message = "Format is not a colon-separated list of alphanumeric strings",
-                        ErrorFix error_fix = ErrorFix::IRRECOVERABLE_VALUE,
-                        const std::string &field = "")
-                : BodySectionError{line, message}, error_fix{error_fix}, field{field} { }
+                        ErrorFix error_fix = ErrorFix::IRRECOVERABLE_VALUE)
+                : BodySectionError{line, message}, error_fix{error_fix} { }
         virtual ~FormatBodyError() override { }
         virtual void apply_visitor(ErrorVisitor &visitor) override { visitor.visit(*this); }
 
         ErrorFix error_fix;
-        const std::string field;
     };
     #pragma db object
     struct SamplesBodyError : public BodySectionError
--- a/inc/vcf/error.hpp
+++ b/inc/vcf/error.hpp
@@ -292,7 +292,7 @@ namespace ebi
                       const std::string &expected_value = "")
                 : BodySectionError{line, message}, error_fix{error_fix}, field{field}, expected_value{expected_value} {
                     if (error_fix == ErrorFix::RECOVERABLE_VALUE && expected_value.empty()) {
-                        throw std::invalid_argument{"An error with recoverable value must provide a non-empty expected value"};
+                        throw std::invalid_argument{"The program had an internal error: An error with recoverable value must provide a non-empty expected value"};
                     }
                 }
         virtual ~InfoBodyError() override { }
--- a/inc/vcf/error.hpp
+++ b/inc/vcf/error.hpp
@@ -124,7 +124,6 @@ namespace ebi
         const std::string message;
         Severity severity;
 
-
       private:
         friend class odb::access;
 
@@ -235,12 +234,14 @@ namespace ebi
     {
         IdBodyError(size_t line = 0,
                     const std::string &message = "ID is not a single dot or a list of strings without semicolons or whitespaces",
+                    ErrorFix error_fix = ErrorFix::IRRECOVERABLE_VALUE,
                     const std::string &field = "")
-                : BodySectionError{line, message}, field{field} { }
+                : BodySectionError{line, message}, error_fix{error_fix}, field{field} { }
         virtual ~IdBodyError() override { }
         virtual void apply_visitor(ErrorVisitor &visitor) override { visitor.visit(*this); }
 
-        std::string field;
+        ErrorFix error_fix;
+        const std::string field;
     };
     #pragma db object
     struct ReferenceAlleleBodyError : public BodySectionError
@@ -274,38 +275,43 @@ namespace ebi
     {
         FilterBodyError(size_t line = 0,
                         const std::string &message = "Filter is not a single dot or a semicolon-separated list of strings",
+                        ErrorFix error_fix = ErrorFix::IRRECOVERABLE_VALUE,
                         const std::string &field = "")
-                : BodySectionError{line, message}, field{field} { }
+                : BodySectionError{line, message}, error_fix{error_fix}, field{field} { }
         virtual ~FilterBodyError() override { }
         virtual void apply_visitor(ErrorVisitor &visitor) override { visitor.visit(*this); }
 
-        std::string field;
+        ErrorFix error_fix;
+        const std::string field;
     };
     #pragma db object
     struct InfoBodyError : public BodySectionError
     {
         InfoBodyError(size_t line = 0,
                       const std::string &message = "Error in info column, in body section",
+                      ErrorFix error_fix = ErrorFix::IRRECOVERABLE_VALUE,
                       const std::string &field = "",
                       const std::string &expected_value = "")
-                : BodySectionError{line, message}, field{field}, expected_value{expected_value} { }
+                : BodySectionError{line, message}, error_fix{error_fix}, field{field}, expected_value{expected_value} {
+                    if (error_fix == ErrorFix::RECOVERABLE_VALUE && expected_value == "") {
+                        throw std::invalid_argument{"An error with recoverable value must provide a non-empty expected value"};
+                    }
+                }
         virtual ~InfoBodyError() override { }
         virtual void apply_visitor(ErrorVisitor &visitor) override { visitor.visit(*this); }
 
-        std::string field;
-        std::string expected_value;    // a few INFO fields expect exact values in a few cases, this would contain that value
+        ErrorFix error_fix;
+        const std::string field;
+        const std::string expected_value;    // a few INFO fields expect exact values in a few cases, this would contain that value
     };
     #pragma db object
     struct FormatBodyError : public BodySectionError
     {
-        FormatBodyError(size_t line = 0,
-                        const std::string &message = "Format is not a colon-separated list of alphanumeric strings",
-                        const std::string &field = "")
-                : BodySectionError{line, message}, field{field} { }
+        using BodySectionError::BodySectionError;
+        FormatBodyError() : FormatBodyError{0} {}
+        FormatBodyError(size_t line) : FormatBodyError{line, "Format is not a colon-separated list of alphanumeric strings"} { }
         virtual ~FormatBodyError() override { }
         virtual void apply_visitor(ErrorVisitor &visitor) override { visitor.visit(*this); }
-
-        std::string field;
     };
     #pragma db object
     struct SamplesBodyError : public BodySectionError
--- a/inc/vcf/error.hpp
+++ b/inc/vcf/error.hpp
@@ -233,11 +233,14 @@ namespace ebi
     #pragma db object
     struct IdBodyError : public BodySectionError
     {
-        using BodySectionError::BodySectionError;
-        IdBodyError() : IdBodyError{0} {}
-        IdBodyError(size_t line) : IdBodyError{line, "ID is not a single dot or a list of strings without semicolons or whitespaces"} { }
+        IdBodyError(size_t line = 0,
+                    const std::string &message = "ID is not a single dot or a list of strings without semicolons or whitespaces",
+                    const std::string &field = "")
+                : BodySectionError{line, message}, field{field} { }
         virtual ~IdBodyError() override { }
         virtual void apply_visitor(ErrorVisitor &visitor) override { visitor.visit(*this); }
+
+        std::string field;
     };
     #pragma db object
     struct ReferenceAlleleBodyError : public BodySectionError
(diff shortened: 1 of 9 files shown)


src/vcf/fixer.cpp
```diff
@@ -128,10 +162,10 @@ namespace ebi
             };
 
             fix_column(info_column_index, string_line, "\t", [&](std::string &info_column) {
-                num_removed_fields = remove_column(info_column, ";", empty_info_column, condition_to_remove_info_field);
+                num_removed_fields = remove_column(info_column, ";", empty_info_column, expected_info_column, condition_to_remove_info_field);
```
jmmut: same as above, split the paths, one for removing a INFO field, another path for replacing a value maybe another for duplicated fields. right now is hard to tell how many fixes this code is performing.

 I'm starting to wonder if this should be expressed explicitly as an enum or something, like `enum {duplication, irrecoverable_value, recoverable_value }`. before moving into that, let's see how this results with several `if`s.

src/vcf/record.cpp
```diff
@@ -318,7 +318,7 @@ namespace ebi
             try {
                 check_no_duplicates(format);
             } catch (const std::invalid_argument &ex) {
-                throw new FormatBodyError{line, "FORMAT must not have duplicate fields", ex.what()};
+                throw new FormatBodyError{line, "FORMAT must not have duplicate fields"};
```
jmmut: I was thinking about this one. we are not fixing this in the debugulator, but could we? I guess it's a bit difficult because we have to look at all the samples to see if the duplicated values match or not, and leave one field in each sample or none, respectively

src/vcf/record.cpp
```diff
@@ -347,45 +347,45 @@ namespace ebi
...
                 }
             }
         } else if (field_key == "SVLEN" && values.size() == alternate_alleles.size()) {
             for (size_t i = 0; i < alternate_alleles.size(); i++) {
                 if (check_alt_not_symbolic(i)) {
                     std::string expected = std::to_string(static_cast<long>(alternate_alleles[i].size()) - static_cast<long>(reference_allele.size()));
                     if (values[i] != expected) {
-                        throw new InfoBodyError{line, "INFO SVLEN=" + field_value + " must be equal to \"length of ALT - length of REF\" for symbolic alternate alleles (expected " + expected + ", found " + values[i] + ")", field_key, expected};
+                        throw new InfoBodyError{line, "INFO SVLEN=" + field_value + " must be equal to \"length of ALT - length of REF\" for symbolic alternate alleles (expected " + expected + ", found " + values[i] + ")", ErrorFix::RECOVERABLE_VALUE, field_key, expected};
```
jmmut: for **non** symbolic alternate alleles

src/vcf/fixer.cpp
```diff
@@ -14,385 +14,582 @@
...
+            }
 
-            // now `it` will point to each SAMPLE column
-            for (auto it = ++first; it != last; ++it) {
-                output << "\t";
-                fix_column(gt_column_index, *it, field_separator, [&](std::string wrong_subfield) {
-                    util::print_container(output, std::string(repeat, empty_subfield), "", subfield_separator, "");
-                });
+            if (fixed_format.empty()) {
+                fixed_format = MISSING_VALUE;         // all format fields were removed, so use missing value
```
Anishka0107: I guess FORMAT does not permit a single dot `.`, we also throw an error in the validator. So in this case, where the Format column must be empty, should we throw an exception or keep it as MISSING_VALUE?

inc/vcf/error.hpp
```diff
@@ -256,32 +260,48 @@ namespace ebi
...
     {
         InfoBodyError(size_t line = 0,
                       const std::string &message = "Error in info column, in body section",
-                      const std::string &field = "")
-                : BodySectionError{line, message}, field{field} {}
+                      ErrorFix error_fix = ErrorFix::IRRECOVERABLE_VALUE,
+                      const std::string &field = "",
+                      const std::string &expected_value = "")
+                : BodySectionError{line, message}, error_fix{error_fix}, field{field}, expected_value{expected_value} {
+                    if (error_fix == ErrorFix::RECOVERABLE_VALUE && expected_value.empty()) {
```
jmmut: This is more a std::logic_error that a invalid argument. if this situation occurs, it is likely to be due to a buggy code. You can also extract this check into a function, as it should be also in FilterBodyError and IdBodyError. I think the prototype could be:
```c++
void assert_recoverable_value_is_provided(ErrorFix error_fix, const std::string &expected_value);
```
And the message could start with "The program had an internal error: "

test/vcf/debugulator_test.cpp
```diff
@@ -314,23 +314,6 @@ namespace ebi
           util::string_split(output.str(), "\t", columns);
           CHECK(columns[8] == "GT:GH");
       }
-      SECTION("Fix FORMAT duplicate field with 1 sample, different values for all")
```
jmmut: This section can be kept, using the macro CHECK_THROWS_AS in the call to the fixer

src/vcf/fixer.cpp
```diff
@@ -14,398 +14,585 @@
...
-    namespace vcf
-    {
-        void Fixer::fix(size_t line_number, std::vector<char> &line, Error &error)
-        {
-            this->line_number = line_number;
-            this->line = &line; // this will be valid until this function returns
-            error.apply_visitor(*this);
-            this->line = nullptr;
-        }
+  namespace vcf
```
cyenyxe: Would it be possible to revert the indentation style to the previous one, as a lot of changes seem to be reported just because of the 4 vs 2 spaces. It is hard for me to see the actual changes, possibly because I was away while this was developed.

Edit: I just saw [@jmmut comment](https://github.com/EBIvariation/vcf-validator/pull/78/commits/cb95994c15fdc2b62ee73598096de8592110d283#r127690791). What do you both think about creating another PR, after this one is merged, to unify the indentation across all the source files in the project?

inc/vcf/fixer.hpp
```diff
@@ -54,13 +56,29 @@ namespace ebi
...
+         * - remove duplicate filter strings - keep the first and remove consequent ones
+         */
         virtual void visit(FilterBodyError &error) override;
+
         /**
-         * fix: remove the info field that caused the error
+         * fix:
+         * - fix duplicates in info field :
+         *     - if even one value differs, remove all the duplicate key fields
+         *     - else, keep one duplicate field and remove all the others
```
Anishka0107: Yes, they'll be kept only if all the values match for the field match. 
e.g. : A=1;A=1;B=2;A=1 gets fixed to A=1;B=2
But A=1;A=2;B=2 gets fixed to B=2

But I guess the comment in the code is not clear? I will update it.

src/vcf/fixer.cpp
```diff
@@ -14,398 +14,585 @@
...
 
-        void Fixer::visit(FileformatError &error)
-        {
-            util::writeline(output, *line);
-            ignored_errors++;
-        }
+    void Fixer::visit(NoMetaDefinitionError &error)
+    {
+        util::writeline(output, *line);
+        ignored_errors++;
```
Anishka0107: Oh yes, sorry I missed it

src/vcf/fixer.cpp
jmmut: I agree on doing a specific PR for silly maintenance. removing the indentation commit from this PR should be trivial

inc/vcf/error.hpp
cyenyxe: Judging by @jmmut comment "it is likely to be due to buggy code", is the idea to abort the program if this occurs because it should never be the case? In that case [an assertion](http://en.cppreference.com/w/cpp/language/static_assert) could be more suitable.

inc/vcf/error.hpp
```diff
@@ -233,11 +232,14 @@ namespace ebi
...
-        IdBodyError() : IdBodyError{0} {}
-        IdBodyError(size_t line) : IdBodyError{line, "ID is not a single dot or a list of strings without semicolons or whitespaces"} { }
+        IdBodyError(size_t line = 0,
+                    const std::string &message = "ID is not a single dot or a list of strings without semicolons or whitespaces",
+                    ErrorFix error_fix = ErrorFix::IRRECOVERABLE_VALUE)
+                : BodySectionError{line, message}, error_fix{error_fix} { }
         virtual ~IdBodyError() override { }
         virtual void apply_visitor(ErrorVisitor &visitor) override { visitor.visit(*this); }
+
+        ErrorFix error_fix;
```
cyenyxe: This attribute is declared in several classes. Would it make sense to derive them from a common parent class that declares said attribute?

inc/vcf/fixer.hpp
```diff
@@ -54,13 +56,30 @@ namespace ebi
...
         virtual void visit(FilterBodyError &error) override;
+
         /**
-         * fix: remove the info field that caused the error
+         * fix:
+         * - fix duplicates in INFO field :
+         *     - if even a single value differs, remove all the duplicate key fields
+         *     - else if the duplicate key has the same value, keep the first one and remove all the others
+         * - if the error is recoverable, i.e., it expected an exact value, then replace the error causing INFO field with the correct value
+         * - if the error is irrecoverable, remove the INFO field that caused the error
```
cyenyxe: Just to be clear on what I meant by capitalization, I meant in plain English. This docstring would look like:

```
/**
 * Fixes duplicates in INFO field.
 * - If even a single value differs, remove all the duplicate key fields
 * - Else if all the duplicate keys have the same value, keep only the first occurrence
 *
 * If the error is recoverable, i.e. it expected an exact value, then replace the error-causing
 * INFO field with the correct value. If the error is irrecoverable, remove the INFO field that 
 * caused the error.
 */
```

src/vcf/fixer.cpp
```diff
@@ -210,10 +243,139 @@ namespace ebi
...
+            return remove_fields(column, separator, is_value_duplicated);
+        }
+
+        size_t Fixer::remove_duplicate_key_value_pairs(const std::string &column,
+                                                       const std::string &separator,
+                                                       const std::string &key_value_separator,
+                                                       const std::string &empty_value)
+        {
+            std::map<std::string, std::string> values;
+            std::vector<std::string> ordered_values;
```
cyenyxe: This variable stores the `ordered_keys`, not the `ordered_values`.

src/vcf/fixer.cpp
```diff
@@ -210,10 +243,139 @@ namespace ebi
...
+                                fixed_samples[it - (first + 1)] += sample_fields[format_fields[ordered_field][0]] + ":";
+                            }
+                        }
+                    } else {
+                        num_removed_duplicates++;
+                    }
+                }
     
+                if (fixed_format.empty()) {
+                    throw std::runtime_error("Could not fix FORMAT duplicate fields, all fields had to be removed, but missing value not permitted");
```
cyenyxe: "Could not fix FORMAT duplicate fields: All fields had to be removed, but missing value is not permitted"

src/vcf/fixer.cpp
```diff
@@ -210,10 +243,139 @@ namespace ebi
...
+        {
+            std::set<std::string> already_present;
+            auto is_value_duplicated = [&](const std::string &value, size_t index) -> bool {
+                bool first_appearance = already_present.insert(value).second;
+                return not first_appearance;
+            };
+            return remove_fields(column, separator, is_value_duplicated);
+        }
+
+        size_t Fixer::remove_duplicate_key_value_pairs(const std::string &column,
```
cyenyxe: This method and the next are quite long. Could they be split in multiple sub-methods?

src/vcf/fixer.cpp
```diff
@@ -210,10 +243,139 @@ namespace ebi
...
+            output << fixed_column;
+            return num_removed_duplicates;
+        }
+
+        size_t Fixer::remove_duplicate_format_sample_pairs(const std::string &string_line) {
+            const size_t format_column_index = 8;
+            size_t num_removed_duplicates = 0;
+
+            using iter = std::vector<std::string>::iterator;
+            fix_columns(format_column_index, -1, string_line, "\t", [&](iter first, iter last) {
```
cyenyxe: I'm trying to think of a way to do this with less data structures.

src/vcf/fixer.cpp
```diff
@@ -325,6 +487,30 @@ namespace ebi
...
+                                     const std::string &separators,
+                                     const std::string &expected_field,
+                                     std::function<bool(const std::string &column, size_t index)> condition_to_replace)
+        {
+            std::vector<std::string> columns;
+            util::string_split(line, separators.c_str(), columns);
+            size_t num_replaced_columns = 0;
+
+            for (size_t j = 0; j < columns.size(); ++j) {
+                if (not condition_to_replace(columns[j], j)) {
```
cyenyxe: I tend to find easier to understand the code when the positive condition is in the if branch, but it may be a personal preference.

inc/vcf/error.hpp
Anishka0107: Yes, but as it was needed in only 4 classes and not all, I didn't extract it😅. Another parameter called `field` is also common to many classes but not derived from the parent class. So in case `error_fix` should be derived from `class Error`, how about deriving `field` too?

src/vcf/fixer.cpp
Anishka0107: Yes, I'll do this and meta_entry.cpp method extraction in the maintenance PR #81 ☺

src/vcf/fixer.cpp
Anishka0107: Ok, I'll do it

src/vcf/fixer.cpp
```diff
@@ -210,10 +243,139 @@ namespace ebi
...
+                return not first_appearance;
+            };
+            return remove_fields(column, separator, is_value_duplicated);
+        }
+
+        size_t Fixer::remove_duplicate_key_value_pairs(const std::string &column,
+                                                       const std::string &separator,
+                                                       const std::string &key_value_separator,
+                                                       const std::string &empty_value)
+        {
```
cyenyxe: I have found a way to improve performance and clarity slightly:

1. `map<string, string> values` stores the first occurrence of a key, so it could be renamed to `first_key_occurrence` for clarity.
2. Turn `set<string> fields_to_remove` into a `vector<bool>` whose values will be true if a field shall be removed.
2.1. The `iterator` variable should be renamed to something more descriptive.
3. Use a numerical index to serially loop through `ordered_keys` and `fields_to_remove`, instead of having to search the whole set every time.

src/vcf/fixer.cpp
```diff
@@ -61,106 +67,135 @@ namespace ebi
...
+            if (error.error_fix != ErrorFix::DUPLICATE_VALUES) {
+                ignore_error();
+                return;
+            }
+        
+            std::cerr << "DEBUG: line " << error.line << ": fixing duplicate FORMAT fields" << std::endl;
+
+            std::string string_line = {line->begin(), line->end()};
+
+            remove_duplicate_format_sample_pairs(string_line);
```
cyenyxe: It would be more clear to wrap this block inside an `if (error.error_fix == ErrorFix::DUPLICATE_VALUES)`, and calling `ignore_error` otherwise. The current way would require more complex code changes if more ErrorFix are supported in the future.

src/vcf/fixer.cpp
Anishka0107: Regarding point 2 and 3, how about using unordered_set instead of set, so that the search is done in O(1) in average case.
Though  the vector<bool> would work the same and is clearly simpler, we would only have to store the duplicate field names in the unordered_set, but in the vector all the fields would be stored.

src/vcf/fixer.cpp
```diff
@@ -210,10 +243,139 @@ namespace ebi
...
+                for (size_t i = 0; i < format_keys.size(); i++) {
+                    format_fields[format_keys[i]].push_back(i);
+                    if (format_fields[format_keys[i]].size() == 1) {
+                        ordered_fields.push_back(format_keys[i]);
+                    }
+                }
+
+                std::set<std::string> fields_to_remove;
+
+                for (auto & format_field : format_fields) {
```
jmmut: this `for` block is building the variable `fields_to_remove`, so it can be extracted into a dedicated function `get_fields_to_remove`

src/vcf/fixer.cpp
```diff
@@ -210,10 +243,139 @@ namespace ebi
...
+                }
+
+                std::string fixed_format;
+                std::vector<std::string> fixed_samples(last - (first + 1));
+
+                for (auto & ordered_field : ordered_fields) {
+                    if (fields_to_remove.find(ordered_field) == fields_to_remove.end()) {
+                        // keep first occurrence, discard the rest if present
+                        fixed_format += ordered_field + ":";
+                        for (auto it = first + 1; it != last; it++) {
```
jmmut: I think it's better to put this for outside the `for (auto & ordered_field : ordered_fields)`. So for each sample, we put all the ordered fields that don't need removal

src/vcf/fixer.cpp
```diff
@@ -210,10 +243,139 @@ namespace ebi
...
+                            std::vector<std::string> sample_fields;
+                            util::string_split(*it, ":", sample_fields);
+                            if (sample_fields.size() > format_fields[ordered_field][0]) {
+                                fixed_samples[it - (first + 1)] += sample_fields[format_fields[ordered_field][0]] + ":";
```
jmmut: this concatenation might be easier using util::print_container, because it accepts the separator, and it handles it so it's not necessary to do the pop_back later

src/vcf/fixer.cpp
cyenyxe: After an offline discussion we have agreed to leave the code mostly as it is right now, but the method shall be split to improve readability.

src/vcf/fixer.cpp
cyenyxe: I think @jmmut and @Anishka0107 discussed a way to optimize the loops in this method a bit, right?

inc/vcf/fixer.hpp
cyenyxe: Could this comment be addressed?

src/vcf/fixer.cpp
```diff
@@ -210,10 +238,145 @@ namespace ebi
...
+
+                for (size_t i = 0; i < format_keys.size(); i++) {
+                    format_fields_indexes[format_keys[i]].push_back(i);
+                    if (format_fields_indexes[format_keys[i]].size() == 1) {
+                        ordered_fields.push_back(format_keys[i]);
+                    }
+                }
+
+                std::set<std::string> fields_to_remove;
+                get_fields_to_remove(format_fields_indexes, first, last, fields_to_remove);
```
cyenyxe: Instead of using an input/output parameter, you could create `fields_to_remove` inside the method and return it, it would feel more natural and there is no need to create it beforehand.

src/vcf/fixer.cpp
```diff
@@ -210,10 +238,145 @@ namespace ebi
...
+                                                       const std::string &empty_value)
+        {
+            std::map<std::string, std::string> first_key_occurrence;
+            std::vector<std::string> ordered_keys;
+            std::set<std::string> fields_to_remove;
+
+            std::vector<std::string> fields;
+            util::string_split(column, separator.c_str(), fields);
+
+            for (auto & field : fields) {
```
cyenyxe: This loop should be extracted to another sub-method similar to the existing `get_fields_to_remove`.

src/vcf/fixer.cpp
```diff
@@ -210,10 +238,149 @@ namespace ebi
...
+            return num_removed_duplicates;
+        }
+
+        size_t Fixer::remove_duplicate_format_sample_pairs(const std::string &string_line) {
+            const size_t format_column_index = 8;
+            size_t num_removed_duplicates = 0;
+
+            using iter = std::vector<std::string>::iterator;
+            fix_columns(format_column_index, -1, string_line, "\t", [&](iter first, iter last) {
+                std::map<std::string, std::vector<size_t>> format_fields_indexes;
```
cyenyxe: Does this really need to be `map<string, vector>`? Only the first position is accessed, so it could be just `map<string, size_t>` instead and the block in lines 309-312 would change a bit. In pseudo-code:

```c++
if (!format_field_indexes.contains(format_keys[i])) {
    format_field_indexes[format_keys[i]] = i;
    ordered_fields.push_back(format_keys[i]);
}
```

src/vcf/fixer.cpp
```diff
@@ -210,10 +238,149 @@ namespace ebi
...
+                }
+
+                if (fixed_format.empty()) {
+                    throw std::runtime_error("Could not fix FORMAT duplicate fields: All fields had to be removed, but missing value is not permitted");
+                }
+                util::print_container(output, fixed_format, "", ":","");
+
+                for (auto it = first + 1; it != last; it++) {
+                    std::vector<std::string> sample_fields;
+                    util::string_split(*it, ":", sample_fields);
```
cyenyxe: Even though in the other method pre-splitting the subfields doesn't give any advantage, in this one the split is done twice, one inside `get_format_fields_to_remove` and another one in this loop.

Considering that errors of this type will be probably systematic across a whole file, and that the number of columns could be in the order of thousands, splitting all those strings twice could really slow down the process.

I suggest the following process:

1. Benchmark how the current code takes to process let's say a file with 10,000 lines and 100 samples. It can be a synthetic file.
2. Change the code to pre-split all the samples' subfields and keep them in a data structure. `get_format_fields_to_remove` would need to receive that data structure as argument.
3. Benchmark the new code against the same file.

Given our previous experience with string splits, I expect to see a difference of at least a 10-15%.

src/vcf/fixer.cpp
```diff
@@ -210,10 +238,149 @@ namespace ebi
...
+            return num_removed_duplicates;
+        }
+
+        std::set<std::string> Fixer::get_format_fields_to_remove(std::map<std::string, std::vector<size_t>> &format_fields_indexes,
+                                                                 std::vector<std::string>::iterator first,
+                                                                 std::vector<std::string>::iterator last)
+        {
+            std::set<std::string> fields_to_remove;
+
+            for (auto it = first + 1; it != last; it++) {
```
cyenyxe: Wouldn't it be more intuitive that `first + 1` was the value of the input parameter, instead of skipping the first element without making it explicit to external callers? If I was to call this method I would expect it to process all the elements from first to last, not to skip the first one.

src/vcf/fixer.cpp
Anishka0107: Yes, we need to use the other positions in the method `get_format_fields_to_remove` (https://github.com/Anishka0107/vcf-validator/blob/debug_new/src/vcf/fixer.cpp#L367)😅
The vector here indicates the indices where the duplicated field is present in the format column. So we have to look at the same indices in the sample columns to check if each column has the same or different values for that field.

PR #81, opened 2017-07-27
Code maintenance

Anishka0107:
For maintenance stuff like extracting big methods, fixing indentation and updating doc comments.

### DIFF
--- a/inc/vcf/fixer.hpp
+++ b/inc/vcf/fixer.hpp
@@ -59,7 +59,7 @@ namespace ebi
         virtual void visit(PositionBodyError &error) override;
 
         /**
-         * Fixes duplicates in ID field. The fix is to keep the first ID and remove consequent ones.
+         * Fixes duplicates in ID field. The fix is to keep the first ID and remove subsequent ones.
          */
         virtual void visit(IdBodyError &error) override;
         virtual void visit(ReferenceAlleleBodyError &error) override;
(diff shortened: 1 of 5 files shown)


src/vcf/meta_entry.cpp
```diff
@@ -103,19 +103,26 @@ namespace ebi
             check_sample(value);
         }
     }
+
+    inline void MetaEntryVisitor::check_key_is_present(std::string const & field, std::string const & key, int key_count) const
+    {
+        if (key_count == 0) {
+            throw new MetaSectionError{entry.line, field + " metadata does not contain a field called '" + key + "'"};
```
cyenyxe: After checking the specification and given that metadata definitions are key-value pairs, I think this string should be written as `field + " metadata does not contain a key called '" + key + "'"`. Even the variable is already called `key`!

inc/vcf/meta_entry_visitor.hpp
```diff
@@ -28,29 +28,35 @@ namespace ebi
...
+    private:
+
+        void check_key_is_present(std::string const & field, std::string const & key, int key_count) const;
+        void check_alt(std::map<std::string, std::string> & value) const;
+        void check_alt_id(std::string const & id_field) const;
+        void check_contig(std::map<std::string, std::string> & value) const;
+        void check_filter(std::map<std::string, std::string> & value) const;
+        void check_filter_id(std::string const & id_field) const;
+        void check_format(std::map<std::string, std::string> & value) const;
+        void check_format_info_number(std::string const & number_field, std::string const & field) const;
```
cyenyxe: `check_format_or_info_number`

inc/vcf/fixer.hpp
```diff
@@ -59,18 +59,16 @@ namespace ebi
         virtual void visit(PositionBodyError &error) override;
 
         /**
-         * fix: 
-         * - remove duplicate IDs, keep the first one and remove consequent ones
+         * Fixes duplicates in ID field. The fix is to keep the first ID and remove consequent ones.
```
Anishka0107: Lol, I always get confused between the two

PR #82, opened 2017-08-04
Simplify build for OS X users (docker based)

Anishka0107:
This PR aims to simplify the build for Mac (not supported as containerized os yet) users by providing a new Dockerfile which runs the tool inside of it. It is not convenient for developers though.

TODO: windows developers/users could use Docker Toolbox to get rid of the restriction of disabling virtualbox. (Fully supports docker, hence write new dockerfile with win base image)

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,9 @@
+.git
+.gitignore
+.dockerignore
+.travis.yml
+build
+docker*
+README.md
+LICENSE*
+install_dependencies.sh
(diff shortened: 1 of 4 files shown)


jmmut:
Can we write here what problems did we have? I honestly don't remember why we decided it was too hard to do the windows build, and couldn't make the mac one work.

cyenyxe:
Running Docker on Windows requires to disable the native Windows virtualization, preventing software like VirtualBox from running.

In both operative systems people could run the binary using Docker Toolbox which doesn't have the same requirements, but the permissions it needs and the fact that the output is still a Linux binary still make it nonviable in some environments; the second issue could be easy to solve though.

Having native, static builds would be the most user-friendly option if we can make it work.

cyenyxe:
Fixed via https://github.com/EBIvariation/vcf-validator/pull/136.

PR #84, opened 2017-09-03
Homogeneous logging and error reporting

Anishka0107:


inc/util/logger.hpp
```diff
@@ -0,0 +1,42 @@
...
+  namespace util
+  {
+    inline void init_boost_loggers()
+    {
+        boost::log::add_console_log
+        (
+            std::clog,
+            boost::log::keywords::format =
+            (
+                boost::log::expressions::stream << "<" << boost::log::trivial::severity << "> " << boost::log::expressions::smessage
```
jmmut: I was a bit confused when I saw an execution.
> `<info> According to the VCF specification, the input file is not valid`

seems odd, compared to 

>  `[info] According to the VCF specification, the input file is not valid`

in my opinion

cyenyxe:
I have run some tests and the trivial logger writes every message to the standard output, no matter what the level is (debug, info, warning, etc). This makes harder to distinguish error messages at first sight, and it will cause conflicts when writing the plain text report, which [also writes to stdout](https://github.com/EBIvariation/vcf-validator/blob/develop/inc/vcf/report_writer.hpp#L42).

We shall discuss how to proceed from here, decide where do we want the logging to be written to and rethink if the trivial logger is enough for our needs.

inc/vcf/summary_report_writer.hpp
```diff
@@ -110,23 +110,31 @@ namespace ebi
     class SummaryReportWriter : public ReportWriter
     {
       public:
-        SummaryReportWriter(std::ostream &out) : out(out) {}
+        SummaryReportWriter(std::string filename)
+        {
+            fout.open(filename, std::ios::app);
```
jmmut: to append might be useful, but everyone will get some kind of error because everyone will assume that the report will be overwritten. I can almost hear this rippling from the future:
> Why it shows the first same error as before? if I just fixed it! :rage:

src/validator_main.cpp
```diff
@@ -153,23 +154,26 @@ namespace
...
-            std::cerr << "Warning, duplicated outputs! will write just once to each output specified by -r/--report" << std::endl;
+            BOOST_LOG_TRIVIAL(warning) << "Duplicated outputs! will write just once to each output specified by -r/--report";
         }
 
         std::vector<std::unique_ptr<ebi::vcf::ReportWriter>> outputs;
 
         for (auto out : outs) {
-            if (out == ebi::vcf::DATABASE) {
+            if (out == ebi::vcf::DATABASE || out == ebi::vcf::TEXT) {
                 auto epoch = std::chrono::system_clock::now().time_since_epoch();
```
jmmut: if we don't extract this out of the `for` loop, files from the same run might get different timestamps

inc/vcf/summary_report_writer.hpp
jmmut: I just saw below that the `main()` ensures that the file doesn't exist. Still, I would use just `std::ios::out`

inc/vcf/fixer.hpp
```diff
@@ -19,11 +19,7 @@
 
 #include <vector>
 #include <set>
-#include <iostream>
-#include <algorithm>
 
-#include "util/stream_utils.hpp"
-#include "util/string_utils.hpp"
```
cyenyxe: Why moving these includes to `fixer.cpp`?

install_dependencies.sh
```diff
@@ -15,7 +15,14 @@ if [ ! -d "odb/odb-2.4.0-x86_64-linux-gnu" ]; then
   wget http://codesynthesis.com/download/odb/2.4/libodb-sqlite-2.4.0.tar.bz2 -O /tmp/libodb-sqlite.tar.bz2
   tar jxvf /tmp/libodb-sqlite.tar.bz2
   cd libodb-sqlite-2.4.0 && ./configure --with-libodb=../libodb-2.4.0 && make && cd ..
+
+  # make easier to find the static libraries
+  cp libodb-2.4.0/odb/.libs/libodb.a .
+  cp libodb-sqlite-2.4.0/odb/sqlite/.libs/libodb-sqlite.a .
```
cyenyxe: If all these are being copied to the working directory, are all the flags still necessary in the `.travis.yml` file?

> cmake -DBUILD_STATIC=1 -DCMAKE_CXX_COMPILER=${CXX_COMPILER} -DCMAKE_C_COMPILER=${C_COMPILER} -DCMAKE_CXX_FLAGS="-Iodb/libodb-2.4.0 -Iodb/libodb-sqlite-2.4.0" -DCMAKE_EXE_LINKER_FLAGS="-Wl,-rpath,odb/libodb-2.4.0/odb/.libs/ -Lodb/libodb-2.4.0/odb/.libs/ -Wl,-rpath,odb/libodb-sqlite-2.4.0/odb/sqlite/.libs/ -Lodb/libodb-sqlite-2.4.0/odb/sqlite/.libs/"  -DODB_PATH=$PWD/odb/ -G Ninja

inc/vcf/summary_report_writer.hpp
```diff
@@ -110,23 +110,31 @@ namespace ebi
     class SummaryReportWriter : public ReportWriter
```
cyenyxe: The documentation is no longer correct: it doesn't write to stdout but to a file.

I am not sure using this class right now is the safest choice, because it is a bit inconsistent in the way it displays messages:

* Warnings are only displayed once
* Meta errors are only displayed once
* The rest of errors are always displayed

Maybe we should convert the `StdoutReportWriter` into a `FileReportWriter` and change the way warnings vs errors are handled later on, once we introduce the real summary errors with counts?

install_dependencies.sh
jmmut: it's not the working directory, it's the main directory for odb

jmmut: the `ls` command below in this script looks like this in travis:
```
$ ./install_dependencies.sh
skipping odb installation: odb directory found:
libodb-2.4.0 libodb-sqlite-2.4.0 odb-2.4.0-x86_64-linux-gnu
libodb.a libodb-sqlite.a
```

jmmut: but you're right, the flags can be simplified to `cmake -DBUILD_STATIC=1 -DCMAKE_CXX_COMPILER=${CXX_COMPILER} -DCMAKE_C_COMPILER=${C_COMPILER} -DCMAKE_CXX_FLAGS="-Iodb/libodb-2.4.0 -Iodb/libodb-sqlite-2.4.0" -DCMAKE_EXE_LINKER_FLAGS="-Wl,-rpath,odb/ -Lodb/  -DODB_PATH=$PWD/odb/ -G Ninja`

... I think :sweat_smile:

inc/vcf/fixer.hpp
Anishka0107: Maybe i moved them because they weren't needed in that hpp file... but are needed in `fixer.cpp`
Should i revert it?

inc/vcf/summary_report_writer.hpp
jmmut: in fact only the NoMetaDefinitionError are written only once in SummaryReportWriter, but I agree with everything else.

inc/vcf/fixer.hpp
jmmut: the current style is keeping most of the includes in hpp files if possible. Exceptions are done in cases of cyclic dependencies and similar.

Nonetheless, moving some includes to cpp files could reduce recompilation times (example: if some a.hpp changed, but it is only used in b.cpp, and c.cpp includes b.hpp, c.cpp will be recompiled without need), although I tinkered with this in the past and IMO **no changes should be made to optimize compilation times without actual measuring** as it is not obvious how the compilations times will change.

inc/vcf/report_writer.hpp
```diff
@@ -35,18 +35,32 @@ namespace ebi
...
     {
         public:
+            FileReportWriter(std::string filename)
+            {
+                fout.open(filename, std::ios::out);
+            }
+
+            ~FileReportWriter()
+            {
+                fout.close();
```
jmmut: although the ofstream can be opened in the constructor and is closed when it's destroyed, I prefer having those explicitly stated like they are now. 

However, don't you think `file` would be a better name than `fout`?

PR #88, opened 2017-09-16
Use Alpine base in docker

### LINKED ISSUE #87: Reduce size of Docker image
Our Docker image is based on Ubuntu Xenial, making it hundreds of MB in size. Reducing it would improve download times and disk usage.

Other well known distributions that can be used as base for Docker images are Alpine and BusyBox. These may not be as straightforward to use as Ubuntu, with more packages needing to be manually installed.

Anishka0107:
Fixes #87

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,6 +1,6 @@
 FROM alpine:3.6
 
-RUN apk upgrade && apk update && apk add \
+RUN apk upgrade && apk add --no-cache \
 g++ \
 make \
 cmake \
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,23 +1,24 @@
-FROM ubuntu:xenial
+FROM alpine:3.6
 
-RUN apt-get update -y && apt-get upgrade -y && apt-get install -y \
-build-essential \
+RUN apk upgrade && apk update && apk add \
+g++ \
+make \
 cmake \
-libboost-dev \
-libboost-filesystem-dev \
-libboost-program-options-dev \
-libboost-regex-dev \
-libsqlite3-dev \
+boost-dev \
+boost-filesystem \
+boost-program_options \
+boost-regex \
+sqlite-dev \
 ragel \
 wget
 
-# Ubuntu-native ODB packages don't seem to work due to a broken ABI, see this thread: http://codesynthesis.com/pipermail/odb-users/2016-May/003277.html
-RUN wget http://www.codesynthesis.com/download/odb/2.4/odb_2.4.0-1_amd64.deb -O /opt/odb_2.4.0-1_amd64.deb
+RUN mkdir /opt
+RUN wget http://www.codesynthesis.com/download/odb/2.4/odb-2.4.0.tar.bz2 -O /opt/odb.tar.bz2
 RUN wget http://codesynthesis.com/download/odb/2.4/libodb-2.4.0.tar.bz2 -O /opt/libodb.tar.bz2
 RUN wget http://codesynthesis.com/download/odb/2.4/libodb-sqlite-2.4.0.tar.bz2 -O /opt/libodb-sqlite.tar.bz2
 
 # ODB compiler
-RUN cd /opt && tar jxvf libodb.tar.bz2 && tar jxvf libodb-sqlite.tar.bz2 && dpkg -i odb_2.4.0-1_amd64.deb
+RUN cd /opt && tar jxvf odb.tar.bz2 && tar jxvf libodb.tar.bz2 && tar jxvf libodb-sqlite.tar.bz2
 RUN cd /opt/libodb-2.4.0 && ./configure && make && make install
 RUN cd /opt/libodb-sqlite-2.4.0 && ./configure --with-libodb=../libodb-2.4.0 && make && make install
 
--- a/docker/entrypoint.sh
+++ b/docker/entrypoint.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/sh
 
 NPROCS=`grep processor </proc/cpuinfo | wc -l`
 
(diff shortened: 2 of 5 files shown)


Anishka0107:
Would you prefer BusyBox?

docker/Dockerfile
```diff
@@ -1,23 +1,24 @@
-FROM ubuntu:xenial
+FROM alpine:3.6
```
cyenyxe: Great, just this has reduced the size of the image from 816 to 516 MB 😄

docker/Dockerfile
```diff
@@ -1,23 +1,24 @@
...
+RUN mkdir /opt
+RUN wget http://www.codesynthesis.com/download/odb/2.4/odb-2.4.0.tar.bz2 -O /opt/odb.tar.bz2
 RUN wget http://codesynthesis.com/download/odb/2.4/libodb-2.4.0.tar.bz2 -O /opt/libodb.tar.bz2
 RUN wget http://codesynthesis.com/download/odb/2.4/libodb-sqlite-2.4.0.tar.bz2 -O /opt/libodb-sqlite.tar.bz2
 
 # ODB compiler
-RUN cd /opt && tar jxvf libodb.tar.bz2 && tar jxvf libodb-sqlite.tar.bz2 && dpkg -i odb_2.4.0-1_amd64.deb
+RUN cd /opt && tar jxvf odb.tar.bz2 && tar jxvf libodb.tar.bz2 && tar jxvf libodb-sqlite.tar.bz2
 RUN cd /opt/libodb-2.4.0 && ./configure && make && make install
 RUN cd /opt/libodb-sqlite-2.4.0 && ./configure --with-libodb=../libodb-2.4.0 && make && make install
```
cyenyxe: I wonder if removing the downloaded packages at the end of the Dockerfile would shrink the image even more... 🤔

docker/Dockerfile
```diff
@@ -1,23 +1,24 @@
...
+g++ \
+make \
 cmake \
-libboost-dev \
-libboost-filesystem-dev \
-libboost-program-options-dev \
-libboost-regex-dev \
-libsqlite3-dev \
+boost-dev \
+boost-filesystem \
```
cyenyxe: Even though this package is downloaded, I am getting these errors when executing `docker run`:

> [ 42%] Building CXX object CMakeFiles/test_validator_v41.dir/test/main_test.cpp.o
[ 44%] Building CXX object CMakeFiles/test_validator_v42.dir/test/main_test.cpp.o
make[2]: *** No rule to make target '/usr/lib/x86_64-linux-gnu/libboost_filesystem.a', needed by 'bin/test_validator_v41'.  Stop.
make[2]: *** Waiting for unfinished jobs....
[ 46%] Building CXX object CMakeFiles/test_validator_v41.dir/test/vcf/parser_v41_test.cpp.o
make[2]: *** No rule to make target '/usr/lib/x86_64-linux-gnu/libboost_filesystem.a', needed by 'bin/test_validator_v42'.  Stop.
make[2]: *** Waiting for unfinished jobs....
[ 48%] Building CXX object CMakeFiles/test_validator_v42.dir/test/vcf/parser_v42_test.cpp.o
make[1]: *** [CMakeFiles/Makefile2:107: CMakeFiles/test_validator_v41.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
make[1]: *** [CMakeFiles/Makefile2:220: CMakeFiles/test_validator_v42.dir/all] Error 2
make: *** [Makefile:95: all] Error 2

Anishka0107: Strange! I am not getting this error. Everything was successful on my system, along with the vcf_validator, vcf_debugulator etc.
I will look into it.

docker/Dockerfile
Anishka0107: And reduced the download size by a great amount 😌

docker/Dockerfile
Anishka0107: I'll try it :)

docker/Dockerfile
cyenyxe: Solved, it was an issue with the CMake cache... 😓

docker/Dockerfile
Anishka0107: Removing wget and the ODB downloads reduced the size by only 2 MBs (from 516mb to 514mb)

jorizci: I think that as the installation of those programs already copies the required files to the respective system directory those /opt directories could be safely deleted altogether. With that you should reduce around 50mb of the resulting size.

But if you really want to reduce layer sizes then the download, compilation and delete should be on the same run command together, ugly as hell, but it's the sad state of the art.

Another thing that could be checked is that after the apk operations you should run a apk clean to remove the package caches, not sure if alpine preserves copies of updated packages (that is another typical thing) and finally if you still have a man directory you can safely delete it.

cyenyxe: > But if you really want to reduce layer sizes then the download, compilation and delete should be on the same run command together, ugly as hell, but it's the sad state of the art.

Do we need to get to that extreme? It is recommended to minimize the number of layers, but we are really really far from the limit. We have already reduced the size quite a lot so I would prefer not to make the Dockerfile horrible to read.

jorizci: The problem it's that to download a image you have to download all the layers, if you do a layer when you create a lot of information and another layer when you remove all that information... It's like yes, your final image weights X but the effective size of all the layer to download is X+Y

That is why the pattern download / configure / compile delete in the same run statement the layer will only be the final diff between the previous layer and the current state of the machine. This doesn't mean that you need to put everything together just the stuff that is related. The alternative are the intermediate builder dockers that are starting to appear but it will probably be too much hassle as they are kind of beta now.

Anishka0107:
Closing this for now. The g++ which comes prepackaged with alpine does not have `--enable-plugin` option, which is required by ODB. This would require compiling gcc from source with that option and then compiling odb.

PR #90, opened 2017-09-19
Use debian base in docker

Anishka0107:
Current size : 621mb
Older size : 816mb

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,6 +1,6 @@
 FROM debian:stretch-slim
 
-RUN apt-get update -y && apt-get upgrade -y && apt-get install -y --no-install-recommends \
+RUN apt-get update -y && apt-get install -y --no-install-recommends \
 g++ \
 make \
 cmake \
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,7 +1,8 @@
-FROM ubuntu:xenial
+FROM debian:stretch-slim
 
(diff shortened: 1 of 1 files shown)


docker/Dockerfile
```diff
@@ -1,25 +1,28 @@
-FROM ubuntu:xenial
+FROM debian:stretch-slim
 
-RUN apt-get update -y && apt-get upgrade -y && apt-get install -y \
-build-essential \
+RUN apt-get update -y && apt-get upgrade -y && apt-get install -y --no-install-recommends \
```
jmmut: is the `apt-get upgrade` necessary? will it install all the updates in the system?

Anishka0107: I guess removing it would be better, as we don't need to upgrade all the packages...

cyenyxe: In fact it seems to be [discouraged](https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/#run).

docker/Dockerfile
```diff
@@ -1,25 +1,28 @@
...
+make \
 cmake \
 libboost-dev \
 libboost-filesystem-dev \
 libboost-program-options-dev \
 libboost-regex-dev \
 libsqlite3-dev \
 ragel \
-wget
+&& apt-get clean \
```
cyenyxe: Adding a comment like `Clean up to reduce layer size` before this line would be useful to make a clear separation of the actions.

PR #91, opened 2017-10-04
Add support for gVCF files

### LINKED ISSUE #86: EVA-152 Support gVCF files
VCF v4.3 introduces in section 5.5 a way to represent unspecified alleles and REF-only blocks (blocks that don't have any variants). A symbolic alternate allele \<*\> is used to represent this unspecified alternate allele, as in:

```
#CHROM   POS    ID   REF   ALT     QUAL   FILTER   INFO       FORMAT     Sample
1        4370   .    G     <*>     .      .        END=4383   GT:DP:GQ   0/0:25:60
1        4384   .    C     <*>     .      .        END=4388   GT:DP:GQ   0/0:25:45
1        4389   .    T     TC,<*>  213.73 .       .           GT:DP:GQ   0/1:23:99
1        4390   .    C     <*>     .      .        END=4390   GT:DP:GQ   0/0:26:0
```

Note that it is possible to specify the \<*\> allele in every variant (see 1:4389) but the specification doesn't indicate it is mandatory.

Anishka0107:
Fixes #86 (maybe)


inc/vcf/string_constants.hpp
```diff
@@ -102,6 +102,9 @@ namespace ebi
     const std::string INS = "INS";
     const std::string INV = "INV";
 
+    // ALT for gVCF
+    const std::string GVCF = "<*>";
```
jmmut: @cyenyxe, what about NON_REFERENCE_ALLELE? NON_REFERENCE?

src/vcf/validate_optional_policy.cpp
```diff
@@ -146,6 +152,36 @@ namespace ebi
         }
     }
 
+    void ValidateOptionalPolicy::check_body_entry_alt_gvcf_end(ParsingState & state, Record const & record) const
+    {
+        if (std::find(record.alternate_alleles.begin(), record.alternate_alleles.end(), GVCF) != record.alternate_alleles.end()
```
jmmut: I'm so tired of this `!= end()` idiom that I'm considering doing a generic function like this:
```
bool contains(collection, element) {
    return std::find(collection.begin(), collection.end(), element) != collection.end();
}
```
@cyenyxe thoughts?

src/vcf/validate_optional_policy.cpp
```diff
@@ -146,6 +152,36 @@ namespace ebi
...
+            && record.info.find(END) == record.info.end()) {
+            throw new AlternateAllelesBodyError{state.n_lines,
```
jmmut: I think this would fit better as INFO error, instead of ALT. The function name should be changed as well probably

src/vcf/validate_optional_policy.cpp
```diff
@@ -146,6 +152,36 @@ namespace ebi
...
+                    "INFO END should be provided, as " + GVCF + " (symbolic alternate allele) is supposed to represent a region"};
+        }
+    }
+
+    void ValidateOptionalPolicy::check_body_entry_alt_gvcf_gt_value(ParsingState & state, Record const & record) const
+    {
+        if (std::find(record.alternate_alleles.begin(), record.alternate_alleles.end(), GVCF) != record.alternate_alleles.end()
```
jmmut: try to be a bit more strict with the line limit of 120 characters, it really improves readability. in fact, I don't recall talking about this, so this might be the first time you hear about this :sweat_smile:

src/vcf/validate_optional_policy.cpp
```diff
@@ -146,6 +152,36 @@ namespace ebi
...
+            && record.format[0] == vcf::GT) {
+            size_t provided_ploidy = state.source->ploidy.get_ploidy(record.chromosome);
+            std::string required_gt_subfield = "0";
+            while (--provided_ploidy) {
+                required_gt_subfield += "/0";
+            }
+            for (auto & sample : record.samples) {
+                if (sample.substr(0, required_gt_subfield.size() + 1) == required_gt_subfield + ":"
+                    || sample == required_gt_subfield) {
```
jmmut: switch this `||` as it will read positions past the end of the string if `sample == required_gt_subfield` is true

test/vcf/optional_policy_test.cpp
```diff
@@ -137,4 +137,215 @@ namespace ebi
...
+                                "chr1",
+                                123456,
+                                { "id123" },
+                                "A",
+                                { vcf::GVCF },
+                                1.0,
+                                { vcf::PASS },
+                                { { vcf::END, "0" } },
+                                { vcf::GT },
+                                { "0|1" },
```
jmmut: this reminds me, is there any `CHECK_NO_THROW( ... "0|0"...`?

Anishka0107: Shouldn't the sole presence of `0|0` generate a warning? 
also, is `0/0` the same as `0|0`?

jmmut: OK, we never really explained a lot of the biological background about VCFs, the vertical line means that the genotypes are phased, which is that it is known to which chromosome the allele belongs. you know that diploid species have 2 copies of each chromosome (humans have 23 pairs) so the genotype reports the allele found in the same position of both copies of a chromosome. So `0/1` is the same as `1/0`, because the `/` means unphased (it's not known which number corresponds to each copy), but `0|1` is **not** the same as `1|0`. In summary, an appearance of `0|0` should prevent raising this warning

Anishka0107: oh ok thanks

inc/vcf/string_constants.hpp
Anishka0107: not done in nexxt commit

src/vcf/validate_optional_policy.cpp
Anishka0107: not done in next commit

src/vcf/validate_optional_policy.cpp
```diff
@@ -146,6 +152,41 @@ namespace ebi
...
+                std::string gt_subfield = sample.substr(0, sample.find(':'));
+                std::vector<std::string> alleles;
+                util::string_split(gt_subfield, "/|", alleles);
+                bool found_in_sample = true;
+                for (auto & allele : alleles) {
+                    if (allele != "0") {
+                        found_in_sample = false;
+                        break;
+                    }
+                }
```
jmmut: what about extracting a function since the substr until here? so that this function is just
```
if (...) {
    for (sample ...) {
        if (sample_has_reference_in_all_alleles(sample)) {
            return;
        }
    }
    throw ...
}
``` 
which reads quite nice

src/vcf/validate_optional_policy.cpp
Anishka0107: And what about this @jmmut @cyenyxe?
Should this be done in a separate PR or in this one itself?

inc/vcf/string_constants.hpp
cyenyxe: I would suggest GVCF_NON_VARIANT_ALLELE because in `record.cpp` we display the following error when dot is listed along with others:

```
The no-alternate alleles symbol (dot) can not be combined with others"};
```

Just to make clear that one is the GVCF definition and the other a more VCF-generic one.

src/vcf/validate_optional_policy.cpp
```diff
@@ -146,6 +152,44 @@ namespace ebi
...
+    {
+        if (std::find(record.alternate_alleles.begin(), record.alternate_alleles.end(), GVCF)
+            != record.alternate_alleles.end() && record.format[0] == vcf::GT) {
+            for (auto & sample : record.samples) {
+                if (sample_has_reference_in_all_alleles(sample)) {
+                    return;
+                }
+            }
+            throw new AlternateAllelesBodyError{state.n_lines,
+                    "At least one sample should contain genotype with all reference alleles, when ALT is " + GVCF
```
cyenyxe: `should report a genotype`

src/vcf/validate_optional_policy.cpp
```diff
@@ -146,6 +152,44 @@ namespace ebi
...
+        return true;
+    }
+
+    void ValidateOptionalPolicy::check_body_entry_info_gvcf_end(ParsingState & state, Record const & record) const
+    {
+        if (std::find(record.alternate_alleles.begin(), record.alternate_alleles.end(), GVCF)
+            != record.alternate_alleles.end() && record.info.find(END) == record.info.end()) {
+            throw new InfoBodyError{state.n_lines,
+                    "INFO END should be provided, as " + GVCF
+                    + " (symbolic alternate allele) is supposed to represent a region"};
```
cyenyxe: For homogeneity with the other GVCF related error:

```
throw new InfoBodyError{state.n_lines,
        "INFO END should be provided when ALT is " + GVCF + " as it is supposed to be a region"};
```

I have avoided calling it a "reference region" to highlight the fact that is a region, plain and simple.

test/input_files/v4.3/passed/passed_body_alt.vcf
```diff
@@ -20,3 +20,7 @@
 1	1800	rs180734498	C	AGT]1:100]	100	PASS	AN=2184;AC=249;AF=0.11	GT:DS:GL	0|0:0.050:-0.13,-0.58,-3.62	0|1:1.000:-2.45,-0.00,-5.00
 1	1900	rs180734498	C	AGT[1:100[	100	PASS	AN=2184;AC=249;AF=0.11	GT:DS:GL	0|0:0.050:-0.13,-0.58,-3.62	0|1:1.000:-2.45,-0.00,-5.00
 1	1900	rs180734498	C	.	100	PASS	AN=2184;AC=249;AF=0.11	GT:DS:GL	0|0:0.050:-0.13,-0.58,-3.62	0|1:1.000:-2.45,-0.00,-5.00
+1	4370	.	G	<*>	.	.	END=4383	GT:DP:GQ:MIN:PL	0/0:25:60:23:0,60,900	0/0:25:45:25:0,42,630
+1	4389	.	T	TC,<*>	213.73	.	.	GT:DP:GQ:PL	0/1:23:99:51,0,36,93,92,86	0/0:24:52:0,52,95,66,95,97
+1	4389	.	C	<*>,<DEL>	.	.	END=4388	GT:DP:GQ	0/0:25:45	0/1:23:99
+1	4390	.	C	*,<*>	.	.	END=4390	GT:DP:GQ	0/0:26:0	0/0:25:60
```
cyenyxe: Are we really accepting this? Is it correct? 🤔

jmmut: > 5. ALT - alternate base(s): Comma separated list of alternate non-reference alleles. These alleles do not have to be called in any of the samples. Options are base Strings made up of the bases `A,C,G,T,N,*`, (case insensitive) or an angle-bracketed ID String (`“<ID>”`) or a breakend replacement string as described in the section on breakends. The `‘*’` allele is reserved to indicate that the allele is missing due to a an overlapping deletion.

src/vcf/validate_optional_policy.cpp
Anishka0107: Test commit added for this: https://github.com/EBIvariation/vcf-validator/pull/91/commits/c426c36c930cd84e2c7803491a9315ede5de29bf.
Will remove it if not needed.

PR #95, opened 2017-11-09
Add summary report for errors & warnings

### LINKED ISSUE #85: EVA-494 include error counts in the report
Currently we have a StdoutReport that writes one line for each error. A more user friendly report would be one where each error type is only reported once with a count of appearances. Optionally with the first error of each type as example.

examples:

change from this:
```
Line 174: ALT metadata ID does not begin with DEL/INS/DUP/INV/CNV
Line 175: ALT metadata ID does not begin with DEL/INS/DUP/INV/CNV
Line 176: ALT metadata ID does not begin with DEL/INS/DUP/INV/CNV
Line 201: Error in INFO metadata
Line 203: Info 'ssID' is not listed in a valid meta-data INFO entry (warning)
According to the VCF v4.1 specification, the input file is not valid
```
to this:
```
ALT metadata ID does not begin with DEL/INS/DUP/INV/CNV. This occurs 3 times, first time in line 174.
Error in INFO metadata. This occurs 3 times, first time in line 201.
Info 'ssID' is not listed in a valid meta-data INFO entry (warning). This occurs 300 times, first time in line 203.
According to the VCF v4.1 specification, the input file is not valid
```
maybe copying the first occurrence (don't focus too much on this if it's too hard, it's not necessary):
```
INFO AC value must be a non-negative integer number. This occurs 300 times, first time in line 14. Original text: AC=-1
```

Anishka0107:
Fixes #85 

Yet to do these:
- [x] Add new tests
- [x] Update comments

### DIFF
--- a/inc/vcf/error.hpp
+++ b/inc/vcf/error.hpp
@@ -114,7 +114,8 @@ namespace ebi
                 : runtime_error{std::string{"Line "} + std::to_string(line) + ": " + message
                                     + (detailed_message != "" ? (". " + detailed_message) : "") + "."},
                   line{line},
-                  message{message} {}
+                  message{message},
+                  detailed_message{detailed_message} {}
 
         virtual ~Error() override { }
 
--- a/inc/vcf/error.hpp
+++ b/inc/vcf/error.hpp
@@ -112,7 +112,7 @@ namespace ebi
 
         Error(size_t line, const std::string &message, const std::string &detailed_message = "")
                 : runtime_error{std::string{"Line "} + std::to_string(line) + ": " + message
-                                    + (detailed_message != "" ? ". " + detailed_message + "." : "")},
+                                    + (detailed_message != "" ? (". " + detailed_message) : "") + "."},
                   line{line},
                   message{message} {}
 
--- a/inc/vcf/error.hpp
+++ b/inc/vcf/error.hpp
@@ -110,8 +110,9 @@ namespace ebi
 
         Error(size_t line) : Error{line, "Error, invalid file."} {}
 
-        Error(size_t line, const std::string &message)
-                : runtime_error{std::string{"Line "} + std::to_string(line) + ": " + message},
+        Error(size_t line, const std::string &message, const std::string &detailed_message = "")
+                : runtime_error{std::string{"Line "} + std::to_string(line) + ": " + message
+                                    + (detailed_message != "" ? ". " + detailed_message + "." : "")},
                   line{line},
                   message{message} {}
 
@@ -122,6 +123,7 @@ namespace ebi
 
         const size_t line;
         const std::string message;
+        const std::string detailed_message;
         Severity severity;
 
       private:
@@ -185,14 +187,10 @@ namespace ebi
     {
       public:
         NoMetaDefinitionError(size_t line,
-                              const std::string &message,
-                              const std::string &column,
-                              const std::string &field)
-                : Error{line, message}, column{column}, field{field} {}
+                              const std::string &message)
+                : Error{line, message} {}
         virtual ~NoMetaDefinitionError() override { }
         virtual void apply_visitor(ErrorVisitor &visitor) override { visitor.visit(*this); }
-        std::string column;
-        std::string field;
       private:
         friend class odb::access;
         NoMetaDefinitionError() {}
@@ -287,6 +285,7 @@ namespace ebi
     {
         InfoBodyError(size_t line = 0,
                       const std::string &message = "Error in info column, in body section",
+                      const std::string &detailed_message = "",
                       ErrorFix error_fix = ErrorFix::IRRECOVERABLE_VALUE,
                       const std::string &field = "",
                       const std::string &expected_value = "")
@@ -329,9 +328,10 @@ namespace ebi
       public:
         SamplesFieldBodyError(size_t line,
                               const std::string &message,
+                              const std::string &detailed_message,
                               const std::string &field,
                               long field_cardinality = -1)
-                : BodySectionError{line, message}, field{field}, field_cardinality{field_cardinality} {
+                : BodySectionError{line, message, detailed_message}, field{field}, field_cardinality{field_cardinality} {
             if (field.empty()) {
                 throw std::invalid_argument{"SamplesFieldBodyError: field should not be an empty string. Use "
                                                     "SamplesBodyError for unknown errors in the samples columns"};
--- a/inc/vcf/summary_report_writer.hpp
+++ b/inc/vcf/summary_report_writer.hpp
@@ -96,7 +96,7 @@ namespace ebi
         {
             for (auto & error_message : summary.error_order) {
                 file << error_message << ". This occurs " << summary.error_summary_report[error_message].appearances
-                     << " time(s), first time in line " << summary.error_summary_report[error_message].first_appearance_line << std::endl;
+                     << " time(s), first time in line " << summary.error_summary_report[error_message].first_appearance_line << "." << std::endl;
             }
         }
     };
--- a/inc/vcf/summary_report_writer.hpp
+++ b/inc/vcf/summary_report_writer.hpp
@@ -25,6 +25,9 @@ namespace ebi
   namespace vcf
   {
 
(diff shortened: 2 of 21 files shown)


src/vcf/validate_optional_policy.cpp
```diff
@@ -99,8 +99,8 @@ namespace ebi
                     if (alleles.size() != ploidy) {
                         throw new SamplesFieldBodyError{
                                 state.n_lines,
-                                "Sample #" + std::to_string(i) + " has " + std::to_string(alleles.size())
-                                        + " allele(s), but " + std::to_string(ploidy) + " were found in others",
+                                "Sample #" + std::to_string(i) + " has " + std::to_string(alleles.size()) + " allele(s)",
+                                "But " + std::to_string(ploidy) + " were found in others",
```
jmmut: Leaving a note to myself to review this case better

inc/vcf/error.hpp
```diff
@@ -185,14 +187,10 @@ namespace ebi
     {
       public:
         NoMetaDefinitionError(size_t line,
-                              const std::string &message,
-                              const std::string &column,
```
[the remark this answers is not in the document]
Anishka0107: The `NoMetaDefinitionError` is ignored by the debugulator -> https://github.com/EBIvariation/vcf-validator/blob/develop/src/vcf/fixer.cpp#L78.
Should i restore it, incase we might need it somewhere else?

inc/vcf/summary_report_writer.hpp
```diff
@@ -40,69 +40,22 @@ namespace ebi
...
-        virtual void visit(MetaSectionError &error) {}
-        virtual void visit(HeaderSectionError &error) {}
-        virtual void visit(BodySectionError &error) {}
-        virtual void visit(NoMetaDefinitionError &error)
-        {
-            if (is_already_reported(error.column, error.field)) {
-                skip = true;
+            if (error_summary_report.find(error_message) == error_summary_report.end()) {
+                error_order.push_back(error_message);
+                error_summary_report.emplace(std::make_pair(error_message, std::make_pair(1, error_line)));
```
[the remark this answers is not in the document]
Anishka0107: Also do we need this separate class caleed `SummaryTracker` now? Can we merge it with `SummaryReportWriter`?

inc/vcf/summary_report_writer.hpp
```diff
@@ -24,88 +24,50 @@ namespace ebi
...
+     * The summary diplays the count(number of times it appears) of the error, and the line number of it's first
+     * occurrence. We distinguish between different types of errors based on their simple error message (which contains
+     * no details). The `error_order` basically maintains the order in which these errors appaer for the first time.
      *
      * Use this class like this:
      * ~~~
-     * if (summary.should_write_report(error)) {
-     *     std::cout << error.what() << " (warning)" << std::endl;
-     * }
+     * summary.add_to_summary(error.message, error.line);
```
jmmut: now that the interface of this class is much simpler, we can remove this comment on how to use this class

inc/vcf/error.hpp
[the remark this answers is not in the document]
jmmut: mh, I see. There's no big reason to keep it or remove it, except that everytime we modify the Error hierarchy, the db reports generated by previous versions become incompatible. However, this PR already modifies it, and I can not see how we would want to change the fix for this error, as it is about a field that is not defined, and we already check those which are defined in the spec.

Let's leave it as it is. If we ever need to bring them back, we put it again.

src/vcf/record.cpp
```diff
@@ -258,24 +258,24 @@ namespace ebi
                         check_field_cardinality(field.second, values, key_values[NUMBER]);
                         check_field_type(values, key_values[TYPE]);
                     } catch (std::shared_ptr<Error> ex) {
-                        std::string message = "INFO " + key_values[ID] + "=" + field.second
-                                + " does not match the meta" + ex->message;
-                        throw new InfoBodyError{line, message, ErrorFix::IRRECOVERABLE_VALUE, key_values[ID]};
+                        std::string message = "INFO " + key_values[ID] + " does not match the meta" + ex->message;
+                        throw new InfoBodyError{line, message, key_values[ID] + "=" + field.second,
```
jmmut: There's another place were you can do this, in record_cache. I tested the summary report and looks great, but not so much for duplicated variants:
> Duplicated variant 0:0:A>G found in lines 34 and 35. This occurs 2 time(s), first time in line 34
Duplicated variant 0:0:A>G found in lines 34 and 36. This occurs 1 time(s), first time in line 36
Duplicated variant 0:0:A>G found in lines 34 and 37. This occurs 1 time(s), first time in line 37
Duplicated variant 0:0:A>G found in lines 34 and 38. This occurs 1 time(s), first time in line 38

As you see, we are including the second appearance (35, 36, 37, 38) in the main message, which ends up in different entries in the map. Ideally this would show as:

> Duplicated variant 0:0:A>G found. This occurs 5 time(s), first time in line 34

the debugulator is able to remove those, so there's no need to tell which lines it was found to the user (those can still be in the detailed message so that the full text report still has the lines).

Anishka0107: ohh sorry i forsgot it

Anishka0107: this is what the errors look like now in the full text report:
```
Line 34: Duplicated variant 0:0:A>G found.
Line 35: Duplicated variant 0:0:A>G found. It occurs in lines 34, 35
Line 36: Duplicated variant 0:0:A>G found. It occurs in lines 34, 36
Line 37: Duplicated variant 0:0:A>G found. It occurs in lines 34, 37
Line 38: Duplicated variant 0:0:A>G found. It occurs in lines 34, 38
```

src/vcf/validate_optional_policy.cpp
cyenyxe: The detailed message could be slightly modified, otherwise the punctuation is going to make it look quite weird. Something like

```
std::to_string(ploidy) + " alleles were found in other samples"
```

inc/vcf/summary_report_writer.hpp
```diff
@@ -24,88 +24,45 @@ namespace ebi
...
   namespace vcf
   {
+
     /**
-     * Class that tells whether an error should be written or skipped.
+     * Stores the count and first line of appearance of an error message
+     */
+    struct ErrorSummary
+    {
+      size_t appearances;
```
cyenyxe: I think the most appropriate word here would be `occurrences` instead of `appearances`?

inc/vcf/summary_report_writer.hpp
```diff
@@ -98,6 +104,8 @@ namespace ebi
                 file << error_message << ". This occurs " << summary.error_summary_report[error_message].occurrences
                      << " time(s), first time in line " << summary.error_summary_report[error_message].first_occurrence_line << "." << std::endl;
             }
+
+            file << report_result << std::endl;
```
jmmut: IMO this would read better as the first line in the summary

src/vcf/fixer.cpp
```diff
@@ -235,7 +235,7 @@ namespace ebi
     void Fixer::visit(DuplicationError &error)
     {
         BOOST_LOG_TRIVIAL(debug) << "Line " << line_number << ": Fixing duplicate: removing variant: "
-                                 << std::string{line->begin(), line->end()};
+                                 << std::string{line->begin(), line->end()-1};
     }
```
Anishka0107: This displayed an extra end line previously. The `line` contained one and another one was added by the logger.

src/vcf/fixer.cpp
```diff
@@ -235,7 +235,7 @@ namespace ebi
     void Fixer::visit(DuplicationError &error)
     {
         BOOST_LOG_TRIVIAL(debug) << "Line " << line_number << ": Fixing duplicate: removing variant: "
-                                 << std::string{line->begin(), line->end()};
+                                 << std::string{line->begin(), line->end()-1};
```
Anishka0107: This displayed an extra end line previously. The `line` contained one and another one was added by the logger.

PR #103, opened 2018-01-10
Fields PL and GL now depend on the ploidy of GT

### LINKED ISSUE #101: check ploidy when validating PL
Hello! Thanks for your work on this.

I think there is an issue when validating `FORMAT/PL` for non-diploid genotypes. Consider the following region on chromosome X:

```
$ bcftools view -H example.vcf
chrX	10980118	rs1265885	C	T	1572	PASS	SNVHPOL=2;MQ=60	GT:GQ:GQX:DP:DPF:AD:ADF:ADR:SB:FT:PL	1/1:159:30:54:0:0,54:0,27:0,27:-88.3:PASS:370,163,0	1/1:135:30:46:0:0,46:0,23:0,23:-77.1:PASS:370,138,0	1:224:30:10:0:0,10:0,6:0,4:-24.3:PASS:231,0

$ ./vcf_validator -i example.vcf 
Reading from input file...
Line 115: Sample #3, PL=231,0 does not match the meta specification Number=G (contains 2 value(s), expected 3)
According to the VCF specification, the input file is not valid
```

I think the number of `PL` values for the male haploid sample should be equal to the the number of alleles ie. 2.

thanks

Jared

Anishka0107:
Fixes #101

The VCF spec defines for `GL` : "In presence of the GT field the same ploidy is expected; without GT field, diploidy is assumed."
It defines `PL` as : "The phred-scaled genotype likelihoods rounded to the closest integer, and otherwise defined precisely as the GL field."

This is probably not the case with other fields whose `Number` is `G` like `GP`, `CNL`, `CNP`, hence not implementing this for those.

### DIFF
--- a/src/vcf/record.cpp
+++ b/src/vcf/record.cpp
@@ -17,6 +17,10 @@
 #include <functional>
 #include <unordered_set>
 
+
+
+#include<iostream>
+
 #include "util/logger.hpp"
 #include "vcf/file_structure.hpp"
 #include "vcf/record.hpp"
@@ -330,7 +334,8 @@ namespace ebi
         auto iterator = tags.find(field_key);
         if (iterator != tags.end()) {
             try {
-                check_field_cardinality(field_key, values, iterator->second.second);
+                if (field_key != GL && field_key != PL && field_key != GP && field_key != CNL && field_key != CNP)
+                    check_field_cardinality(field_key, values, iterator->second.second);
                 check_field_type(values, iterator->second.first);
             } catch (std::shared_ptr<Error> ex) {
                 raise(std::make_shared<Error>(line, field_key + " does not match the" + ex->message));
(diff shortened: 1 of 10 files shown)


src/vcf/record.cpp
```diff
@@ -536,6 +554,19 @@ namespace ebi
                     throw new SamplesFieldBodyError{line, message + " does not lie in the interval [0,1]", "", field_key};
                 }
             }
+        } else if (field_key == GL || field_key == PL) {
```
jmmut: Add GP as well

test/input_files/v4.3/failed/failed_meta_format_017.vcf
```diff
@@ -1,5 +0,0 @@
-##fileformat=VCFv4.3
```
jmmut: for 4.3 I think these tests were correct

test/vcf/metaentry_test.cpp
```diff
@@ -719,14 +711,6 @@ namespace ebi
                             }),
                             vcf::MetaSectionError* );
                                 
-            CHECK_THROWS_AS( (vcf::MetaEntry {
```
jmmut: keep this and GL  and GP for 4.3 only

Anishka0107:
Just adding a commit with partial work.
**Not yet ready for review!!**

(Doesn't work and also has ugly code)

Issue #105, opened 2018-01-31
Add newer versions of Clang to Travis CI build


cyenyxe:
The versions of Clang used in the Travis CI build are not ancient but they are not new either. New versions include more checks and would also support features from the latest versions of the C++ language.

Add new environment configurations to the `.travis.yml` file to use these more modern versions of the compiler, and fix any compile/test issues that could arise.

PR #110, opened 2018-02-01
clang 3.8 and 3.9 added

InderKaur:
New versions of clang are added to .travis.yml file. #105

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,17 +26,17 @@ matrix:
             - libboost-log1.55-dev
             - ninja-build
 
-    # Clang 3.6
+    # Clang 4.0
     - compiler: clang
-      env: C_COMPILER=clang-3.6 CXX_COMPILER=clang++-3.6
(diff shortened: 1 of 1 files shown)


jmmut:
why both 3.8 and 3.9? each compiler we add makes the build slower, so we have to justify each addition. I didn't have time to check which version is used by default in newer linuxes and newer OSXs.

cyenyxe:
Clang 4 and 5 are now supported in Travis, see https://github.com/travis-ci/apt-package-whitelist/pull/3297 and https://github.com/travis-ci/apt-package-whitelist/pull/4034

What I think we should do is to deprecate some of the oldest versions by checking what is the last Ubuntu LTS onwards, and add these more modern ones to the build. I would also keep at least one 3.x version.

InderKaur:
I have done the requested changes. please have a look

cyenyxe:
Any particular reason why you have left the oldest Clang 3.x version?

cyenyxe:
Closed due to lack of activity.

Issue #104, opened 2018-01-31
Warn if the input VCF is compressed


cyenyxe:
If a compressed file or stream is provided using the `-i` parameter, regular errors are reported, but there's no clue that the input should be uncompressed.

A solution may involve reading magic numbers or extensions, although extra care must be taken in the case of reading from stdin.

Issue #107, opened 2018-01-31
Read compressed files directly


cyenyxe:
Currently, if the input VCF is compressed, it is the user's responsibility to decompress in into a file or using a pipe. Allowing to read from compressed files would make the validator faster to work with. [An example of how to do this using Boost](https://stackoverflow.com/a/6421029) is available in Stack Overflow.

Like any other dependencies in the project, it must be possible to compile the new dependency in static mode.

jmmut:
done in #117

PR #111, opened 2018-02-06
display warning if the input VCF is compressed

srbcheema1:
This PR adds the feature to display warning on providing compressed VCF file as STDIN or using -i parameter.
In reference to the issue #104


srbcheema1:
@cyenyxe @jmmut  
Added a function to check if a file is compressed or not 
This function prints a warning message if the file is compressed. 
Please review

src/vcf/validator.cpp
```diff
@@ -132,6 +132,65 @@ namespace ebi
...
+    bool is_compressed_file(std::istream &input,
+                            const std::string &source)
+    {
+        std::string message = "Input file is a compressed file ";
+
+        if (source != ebi::vcf::STDIN) {
+            boost::filesystem::path source_name(source);
+            std::string file_extension = source_name.extension().string();
+
+            if (file_extension == ".zip") {
```
Anishka0107: You could extract these strings ("zip", "tar", etc) to the file `string_constants.hpp`.

Also the multiple `if` statements could be merged into one single statement.

src/vcf/validator.cpp
```diff
@@ -132,6 +132,65 @@ namespace ebi
...
+                BOOST_LOG_TRIVIAL(warning) << message << "(zip)";
+                return true;
+            }
```
jmmut: this block is repeated several times. it should be extracted to another function

src/vcf/validator.cpp
```diff
@@ -132,6 +132,65 @@ namespace ebi
...
+
+        const unsigned char zip[4] = { 80, 75, 3, 4 };
+        const unsigned char tar_gz[2] = { 31, 139 };
+        const unsigned char tar_xz[5] = { 253, 55, 122, 88, 90 };
+        const unsigned char tar_z[5] = { 31, 157 };
+
+        if (std::equal(zip, zip + sizeof(zip), magic)) {
+            BOOST_LOG_TRIVIAL(warning) << message << "(zip)";
+            return true;
+        }
```
jmmut: this should also be extracted to a function

src/vcf/validator.cpp
```diff
@@ -132,6 +132,65 @@ namespace ebi
...
+                BOOST_LOG_TRIVIAL(warning) << message << "(tar.xz)";
+                return true;
+            }
+            else if (file_extension == ".rar") {
+                BOOST_LOG_TRIVIAL(warning) << message << "(rar)";
+                return true;
+            }
+        }
+
+        unsigned char magic[10];
```
jmmut: until here, the check is done with the extension, and from here to the end it checks the magic numbers. both parts should be extracted to functions, and `is_compressed_file` therefore should contain just calls to those two functions.

src/validator_main.cpp
```diff
@@ -222,13 +222,19 @@ int main(int argc, char** argv)
...
+                return 1;
+            }
             is_valid = ebi::vcf::is_valid_vcf_file(std::cin, path, validationLevel, ploidy, outputs);
         } else {
             BOOST_LOG_TRIVIAL(info) << "Reading from input file...";
             std::ifstream input{path};
             if (!input) {
                 throw std::runtime_error{"Couldn't open file " + path};
             } else {
+                if (ebi::vcf::is_compressed_file(input, path)) {
```
jmmut: note that `is_compressed_file` is consuming data from the stream, and in case the input is not compressed, the parser won't receive the first bits, thus giving false reports for any input file.

jmmut: look at valitador.hpp, there is a part of the code that copies the first line, maybe there is a better place to include this check. also it wouldn't be necessary to write the check twice, as in here.

src/vcf/validator.cpp
```diff
@@ -132,6 +132,65 @@ namespace ebi
         }
     }
 
+    bool is_compressed_file(std::istream &input,
+                            const std::string &source)
+    {
+        std::string message = "Input file is a compressed file ";
```
jmmut: "Input file should not be compressed (detected " + compression_type + " compression)"

src/vcf/validator.cpp
```diff
@@ -132,6 +132,65 @@ namespace ebi
...
+            }
+        }
+
+        unsigned char magic[10];
+        input.read((char*)magic, sizeof(magic));
+
+        const unsigned char zip[4] = { 80, 75, 3, 4 };
+        const unsigned char tar_gz[2] = { 31, 139 };
+        const unsigned char tar_xz[5] = { 253, 55, 122, 88, 90 };
+        const unsigned char tar_z[5] = { 31, 157 };
```
jmmut: this is optional, but I don't like having to put the count of elements. this always ends up slowing modifications a lot. to change this, find alternate ways to specify the array, or use a std::vector.

src/vcf/validator.cpp
srbcheema1: sure.
I will move the strings to `inc/vcf/string_constants.hpp` file.
Should I move the warning message strings too ?
I was checking in `validator_main.cpp` few strings(used in info messages) were used there only.

src/vcf/validator.cpp
srbcheema1: I will merge these `if-else-if` statements into one. and also make a seperate function for this extension checking

src/vcf/validator.cpp
srbcheema1: sure .
I will extract this into a seperate function

src/validator_main.cpp
srbcheema1: `is_compressed_file` is consuming data
I think we may use `seekg(0)` for that.

src/vcf/validator.cpp
srbcheema1: Should I extract `Input file should not be compressed (detected` to `inc/vcf/string_constants.hpp` ? it will be used twice . once in function to check with extensions and one in function to check with magic numbers

src/vcf/validator.cpp
srbcheema1: I have made a vector of vectors for it and we can easily use for each loop for checking them all

src/vcf/validator.cpp
jmmut: we are not extracting messages, only words that can actually provoke bugs if misspelled. extracting the messages only makes sense for i18n, which we are not likely to do in the short to mid term

src/vcf/validator.cpp
jmmut: just put the constant at the beginning of this file, or even better, use a function that receives the type of compression and builds the message and does the logging

srbcheema1: building a function looks good :grinning:

srbcheema1:
I have made the requested the changes to `validator.cpp` file and also extracted strings to `string_constants.hpp` file.
Tests are yet to be made.
Have a look at the changes. Once the changes get approved I will squash the commits into one.

src/validator_main.cpp
srbcheema1: Check is added twice due to the reason that once the input is in form of STDIN and once we are reading from file .
Even the `is_valid_vcf_file` check is used twice in this code.

src/vcf/validator.cpp
```diff
@@ -132,6 +132,63 @@ namespace ebi
...
+            }
+        }
+        return false;
+    }
+
+    bool is_compressed_file(std::istream &input,
+                            const std::string &source)
+    {
+        if (source != ebi::vcf::STDIN) {
+            if (compare_extension(input, source)) {
```
Anishka0107: You could probably group these two if's into one using AND `&&`

inc/vcf/string_constants.hpp
```diff
@@ -174,6 +174,14 @@ namespace ebi
     const std::string PQ = "PQ";
     const std::string PS = "PS";
 
+    //Extensions of compressed files
```
Anishka0107: Add a space after the comment begins `//`
I would prefer this line as `// Compressed file extensions`

src/validator_main.cpp
```diff
@@ -222,13 +222,19 @@ int main(int argc, char** argv)
 
         if (path == ebi::vcf::STDIN) {
             BOOST_LOG_TRIVIAL(info) << "Reading from standard input...";
+            if (ebi::vcf::is_compressed_file(std::cin, path)) {
+                return 1;
+            }
             is_valid = ebi::vcf::is_valid_vcf_file(std::cin, path, validationLevel, ploidy, outputs);
```
Anishka0107: You could avoid adding these 4 lines twice : Lines 225-228 and 235-238.

srbcheema1:
@Anishka0107 I have made the requested changes.
Please have a look.

src/vcf/validator.cpp
```diff
@@ -132,12 +132,66 @@ namespace ebi
         }
     }
 
+    void compressed_file_warning(std::string const & file_extension)
+    {
+        BOOST_LOG_TRIVIAL(warning) << "Input file should not be compressed (detected " << file_extension
+            << " compression)";
+    }
+
+    bool compare_extension(std::istream &input,
```
jmmut: the parameter input is not used, remove it

src/vcf/validator.cpp
```diff
@@ -132,12 +132,66 @@ namespace ebi
...
+                return true;
+            }
+        }
+        return false;
+    }
+
+    bool is_compressed_file(std::istream &input,
+                            const std::string &source)
+    {
+        if (source != ebi::vcf::STDIN && compare_extension(input, source)) {
```
jmmut: another style thing, put the most general functions first, and right below them the functions they use. e.g.:
```
f1() { f2(); f3();}
f2() { }
f3() { } 
f4() { f5() }
f5() { }
``` 
in this case the order should be `is_compressed_file`, `compare_extension`, `compressed_file_warning`,  `compare_magic_num`.

also, I think changing `compare_extension` to `is_compressed_extension` and `compare_magic_num` to `is_compressed_magic_num` express better what the function returns.

jmmut: what's more, all these functions should be right below is_valid_vcf_file.

srbcheema1: According to this, I think `compressed_file_warning` should come below `compare_magic_num`.

jmmut: well, it would be with breadth first graph navigation, but we do deep first. In other words, put the used functions after the first usage

srbcheema1:
@jmmut  I want to discuss something regarding #107 issue . 
there are tools in boost iostream library to read bz2 and gzip compression formats . 
what about returning compression type from `is_compressed_file` and this would help us in reading the compressed files .

jmmut:
I wouldn't worry about #107. For now, just make simple code that does one thing, and when we start that issue, we'll deal with any needed refactoring.

Really. One of the biggest lessons I learnt in recent years is how important is to make simple code. One problem at a time. Usually trying to be too clever makes things worse.

 Let's finish this issue first, I'd say only the tests are left. Let me know if you have any doubt on how to design the tests.

srbcheema1:
I can figure out that i need to add testcase to `test/vcf/parser_v*_.test.cpp` file.
Am I going in right direction ?

jmmut:
There are 2 types of tests we can do here.

- integration tests
- unit test

the integration ones are almost done. Just rename the compressed files you added from ` test/input_files/v4.3/compressed_files/failed_body_alt_000.vcf.*` to `test/input_files/v4.3/failed/compressed.vcf.*`. that way, the test/vcf/parser_v43_test.cpp will try to do a whole validation on those files, and expect them to fail. You don't need to modify test/vcf/parser_v43_test.cpp.

the unit tests should use the functions and check that they return `true` and `false` appropiately. For that, you'll have to move the declarations to validator.hpp (I didn't realise they were in the cpp until now). You can add a simple test file, taking as example ploidy_test.cpp, for instance. Testing the extension function is easy. To test the magic number, you might need to write into a stringstream those bytes, but should be easy as well.

srbcheema1:
I have made a file `tests/vcf/compressed_files_test.cpp` . In that i have added tests to test `is_compressed_magic_num()` and `is_compressed_extension` 
I have renamed those compressed files . 
I haven't yet moved those files into failed folder. By keeping them in seperate folder it was easy for me to write tests for files that return `true` on `is_compressed_*()`

test/vcf/compressed_file_test.cpp
```diff
@@ -0,0 +1,53 @@
...
+      SECTION("Compressed file streams")
+      {
+          auto folder = boost::filesystem::path("test/input_files/v4.3/compressed_files");
+          std::vector<boost::filesystem::path> v;
+          copy(boost::filesystem::directory_iterator(folder), boost::filesystem::directory_iterator(), back_inserter(v));
+
+          for (auto path : v)
+          {
+              std::ifstream input{path.string()};
+              CHECK(vcf::is_compressed_magic_num(input));
```
jmmut: This are the positive cases. Please add test cases for the negative ones as well. Non compressed extensions and non-compressed filestreams, with a special check that the seekg is working as expected.

srbcheema1: i think `seekg()` is automatically tested in `is_valid_vcf_file` test for non compressed files in `parser_v4*` tests

srbcheema1: if `seekg()` behaves abnornally then for sure our `is_valid_vcf_file` test will fail for non compressedfiles passed files(as they will return `false` on `is_valid_vcf_file` ). And we will come to know something is going wrong.

jmmut: that's true, but a specific unit test is also desirable. If only the integration tests detect that, you will have dozens of them failing with strange messages, and not a single unit test failing. The point is not so much to see that current code works (which we know it does) but to protect us from future breaking changes, and being able to identify the problem quickly.

srbcheema1:
I have added the tests for non compressed files also. I have also added a special test to check working of `seekg()`.

srbcheema1:
@jmmut also please tell me if I should squash the commits. I will squash similar commits into one once you approve the changes.

srbcheema1:
@jmmut there was a minor issue with seekg . It was behaving abnormally for files with number of characters less than 9 characters.
I fixed it as explained in https://stackoverflow.com/questions/48979934/seekg-not-behaving-as-expected
using input.clear()
please have a look at it.
I am looking forward to get this PR merged.

src/validator_main.cpp
srbcheema1: were you talking this about https://github.com/EBIvariation/vcf-validator/blob/abbc6fbe2fac9769d8c7c82daa756a66ea2ceba1/src/vcf/validator.cpp#L142
in https://github.com/EBIvariation/vcf-validator/pull/111#discussion_r166876466 "the code that copies first line."
I am thinking to use this `line` in `is_compressed_file` instead of `input` keeping this thing in mind that `seekg` may behave abnormally in future.

jmmut:
hm, that is exactly the issue I was afraid of. what's more, it's not fixed. if you try this:
```
./builds/debug/bin/vcf_validator -i test/input_files/v4.3/passed/complexfile_passed_000.vcf
```
it will say it's valid, but if you pass it through a pipe instead, it will say it's not valid:
```
cat test/input_files/v4.3/passed/complexfile_passed_000.vcf | ./builds/debug/bin/vcf_validator
```
I suggest leaving the approach of seekg and using a copy of the first line as done in https://github.com/EBIvariation/vcf-validator/blob/develop/src/vcf/validator.cpp#L153. there you can see that the first line is read, and it's used for detecting the version in line 145, and to start the validation in line 153.

srbcheema1:
tests were passing.  I will check that out.
I too was thinking to use first line https://github.com/EBIvariation/vcf-validator/pull/111#discussion_r170569585
earlier when you mentioned about that in https://github.com/EBIvariation/vcf-validator/pull/111#discussion_r166876466 I searched such function in `validator.hpp`(maybe you made typo while  writing`validator.cpp`) but was unable to find one. by that time i got approach of seekg as i have used it in filestreams but i was doubtful for using it in std::cin .

jmmut:
yes, that's what I meant, sorry for not being clear enough. I'll try to be more specific in the future, and please ask any doubt you have.

srbcheema1:
I have tested it using these commands 

`./build/bin/vcf_validator < test/input_files/v4.3/passed/complexfile_passed_000.vcf`

and this one

`./build/bin/vcf_validator -i test/input_files/v4.3/passed/complexfile_passed_000.vcf`


I use `command < filename` more than `cat filename| command`

maybe thats why these things went unnoticed . 
moreover our testcases are not capable to check this bug out .

jmmut:
Didn't know `cat file |` and ` < file` could behave so differently. I used `cat file | ` because if the file is compressed the way to use it is simpler and more similar, as ` zcat file.gz |`. 

Agree that the tests (as they are right now) can not check this. This is another reason to go away from stream handling.

srbcheema1:
hmm strange it is. Till now i thought both the ways similar to giving standard input

inc/vcf/validator.hpp
```diff
@@ -187,12 +187,12 @@ namespace ebi
                            Ploidy ploidy,
                            std::vector<std::unique_ptr<ebi::vcf::ReportWriter>> &outputs);
 
-    bool is_compressed_file(std::istream &input,
-                            const std::string &source);
+    bool is_compressed_file(const std::string &source,
+                            std::vector<char> &line);
```
jmmut: make this const

inc/vcf/validator.hpp
```diff
@@ -183,6 +186,14 @@ namespace ebi
                            ValidationLevel validationLevel,
                            Ploidy ploidy,
                            std::vector<std::unique_ptr<ebi::vcf::ReportWriter>> &outputs);
+
+    bool is_compressed_file(const std::string &source,
+                            std::vector<char> &line);
+
+    bool is_compressed_extension(std::string const & source);
+
+    bool is_compressed_magic_num(std::vector<char> &line);
```
jmmut: make this const as well

src/vcf/validator.cpp
```diff
@@ -141,6 +143,9 @@ namespace ebi
         std::vector<char> line;
         ebi::util::readline(input, line);
         ebi::vcf::Version version;
+        if (ebi::vcf::is_compressed_file(sourceName, line)) {
+            return false;
```
jmmut: We talked about how this will be shown at the end to the user:
> [info] Reading from input file...
> [warning] Input file should not be compressed (detected .gz compression)
> [info] According to the VCF specification, the input file is not valid

And we are not sure the last line is very clear. It's not really true that the VCF is not valid, it's just in a format we can't accept right now. So another change is needed. Here, instead of returning false (which means it's not valid) just throw a `std::invalid_argument`. If you see, [the main function](https://github.com/EBIvariation/vcf-validator/blob/develop/src/validator_main.cpp#L243) is prepared so that it logs an error if an exception is thrown.

Also another couple of changes will be needed, in `compressed_file_warning` you can just say 
`BOOST_LOG_TRIVIAL(warning) << "detected " << file_extension << " compression";` so that it's not redundant, and change the tests so that they check that exceptions are thrown, instead of returning false. 

With this change, the output should be:
> [info] Reading from input file...
> [warning] detected .gz compression
> [error] Input file should not be compressed

Which is much more clear IMO.

srbcheema1:
@jmmut I have made the requested changes.
Please have a look at it.

Issue #106, opened 2018-01-31
Add newer versions of GCC to the Travis CI build


cyenyxe:
The versions of GCC used in the Travis CI build are not ancient but they are not new either. New versions include more checks and would also support features from the latest versions of the C++ language.

Add new environment configurations to the `.travis.yml` file to use these more modern versions of the compiler, and fix any compile/test issues that could arise.

jmmut:
given that newer versions were added but the gcc 7 was agreed not to be included because it was too slow, we can close this.

PR #113, opened 2018-02-14
Adds gcc 7 to travis CI

srbcheema1:
This PR adds gcc 7 to Travis file. In reference to the issue #106

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -134,6 +134,24 @@ matrix:
             - libboost-log1.55-dev
             - ninja-build
 
+    # GCC 7
+    - compiler: gcc
+      env: C_COMPILER=gcc-7 CXX_COMPILER=g++-7
+      addons:
+        apt:
+          sources:
+            - boost-latest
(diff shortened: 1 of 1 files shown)


cyenyxe:
It seems the Travis build timed out. Could you please re-run to make sure it wasn't due to an outage in their infrastructure?

srbcheema1:
I too was looking at that thing timed-out thing. I am thinking that gcc 7.3 is not fully stable yet.

cyenyxe:
So it has taken more than 20 minutes to run... I am afraid that will make the Travis builds a lot more unstable, meaning they may fail arbitrarily due to timeouts. We appreciate the work done but we can't merge this unless gcc7 gets faster 😞

srbcheema1:
@cyenyxe I recently saw that the time for gcc 7 build has decreased a lot on Travis. It got down to 15 mins some days back. And now it builds in 13 mins on an average :), which is good I think. I triggered the build thrice and noted an average of 13 mins.
So is it okay to reopen this PR?

Issue #102, opened 2017-12-15
v0.6 not properly validating PL field in VCF 4.1


garysaunders:
PL=. returning an error in validation of VCF 4.1 file.

Although 4.3 defines PL as must be Number=G, this is not a requirement of VCF 4.1 or 4.2.

PR #114, opened 2018-02-22
EVA-1022 Different ploidies should be allowed for a given variant 

jmmut:
following #101 and #102, I took the commits from #103 and finished them:
- Now, in any given row, each sample can have different ploidy.
- As the ploidy affects the cardinality of fields like PL, with this PR each sample can have different cardinality in those fields.
- I kept PL Number=G even in 4.1 and 4.2, because the problem (different samples needed different field cardinality) is fixed by the line above this one.
- Updated tests and documentation accordingly.


src/vcf/record.cpp
```diff
@@ -257,7 +259,7 @@ namespace ebi
                 if (key_values[ID] == field.first) {
                     found_in_meta = true;
                     try {
-                        check_field_cardinality(field.second, values, key_values[NUMBER]);
+                        check_info_field_cardinality(field.first, values, key_values[NUMBER]);
```
cyenyxe: Were we missing all these errors before? Are new tests added to check it?

src/vcf/record.cpp
```diff
@@ -620,28 +661,50 @@ namespace ebi
...
+    }
+
+    void Record::check_sample_field_cardinality(std::string const &field, std::vector<std::string> const &values,
+                                                std::string const &number, size_t ploidy, bool &valid,
+                                                long &cardinality) const
     {
-        long expected;
-        if(not is_valid_cardinality(number, alternate_alleles.size(), expected)) {
-            raise(std::make_shared<Error>(line, field + " meta specification Number=" + number + " is not one of [A, R, G, ., <non-negative number>]"));
+        valid = is_valid_cardinality(number, alternate_alleles.size(), ploidy, cardinality);
```
cyenyxe: `valid` doesn't seem to be used by any of the methods that call this one, so it could be a local variable. In my opinion the method call was clear enough, because it expressed what the validity applied to (cardinality), whereas the variable name is slightly more ambiguous.

src/vcf/record.cpp
```diff
@@ -620,28 +661,50 @@ namespace ebi
...
-            // if number=".", then `expected` was set to -1, and it should always match, letting `number_matches` as true
+            // if number=".", then `cardinality` was set to -1, and it should always match, letting `number_matches` as true
         }
 
         if (!number_matches) {
-            raise(std::make_shared<Error>(line, " specification Number=" + number + " (expected " + std::to_string(expected) + " value(s))"));
+            std::string detailed_message;
+            if (number == G) {
+                detailed_message = ". It must derive its number of values from the ploidy of GT (if present), or "
+                        "assume a diploidy. Contains " + std::to_string(values.size()) + " value(s), expected "
```
cyenyxe: `assume diploidy`, without `a`

src/vcf/record.cpp
```diff
@@ -620,28 +661,50 @@ namespace ebi
...
+                        + std::to_string(cardinality) + " (derived from ploidy " + std::to_string(ploidy) + ")";
+            }
+            raise(std::make_shared<Error>(line, " specification Number=" + number + " (expected "
+                    + std::to_string(cardinality) + " value(s))", detailed_message));
```
cyenyxe: I think it would be more clear if the first part of the message was set first as a separate variable, then the detail in the other one. What does this concatenate if `detailed_message` is undefined?

test/input_files/v4.2/passed/passed_ploidy_001.vcf
```diff
@@ -0,0 +1,6 @@
+##fileformat=VCFv4.1
+#CHROM	POS	ID	REF	ALT	QUAL	FILTER	INFO	FORMAT	HG00096	HG00097
+1	61462	.	T	A	100	PASS	.	GT:PL	0|0:1,2,3	1|1:4,5,6
+2	61462	.	T	A,C	100	PASS	.	GT:PL	0|0:1,2,3,4,5,6	0|1|1:10,11,12,13,14,15,16,17,18,19
```
cyenyxe: Maybe test genotype 0|1|2 to ensure that the index numbers don't have any effect?

src/vcf/record.cpp
jmmut: this `field.first` was used just for the human readable part in the report. instead of something like "Field GT was invalid as 0/0" something like "Field 0/0 was invalid as 0/0". I don't think it mattered a lot, but I will review this as I think that parameter is not needed inside `check_info_field_cardinality` anymore

src/vcf/record.cpp
jmmut: it's used in check_sample_subfields_cardinality_type

src/vcf/record.cpp
jmmut: agree, but detailed_message can not be undefined, it's just an empty string.

src/vcf/record.cpp
jmmut: yep, it was not used anyway

src/vcf/record.cpp
jmmut: I don't think we actually needed that parameter, so I could revert this back

src/vcf/record.cpp
jmmut: and the files that check this (wrong cardinality in INFO fields defined in the header) are test/input_files/v4.3/failed/failed_body_info_029.vcf, 30 and 31.

Issue #115, opened 2018-03-01
Log what reports will be written


jmmut:
Right now the workflow to use the validator is:
- validator (output says ok or not ok)
- `ls -l`
- less newest-report.txt

it would be helpful to write in the log (info level) the file names of each report that will be written. This could be done in the `main` function, after the parameter `-r` is parsed.

PR #116, opened 2018-03-01
Adds log to display filename

srbcheema1:
This PR is for the issue #115
I have added a message to display the filename of output file

### DIFF
--- a/src/validator_main.cpp
+++ b/src/validator_main.cpp
@@ -187,11 +187,11 @@ int main(int argc, char** argv)
         }
 
         std::string report_result = "According to the VCF specification, the input file is " + std::string(is_valid ? "" : "not ") + "valid";
-        BOOST_LOG_TRIVIAL(info) << report_result;
         for (auto & output : outputs) {
-            BOOST_LOG_TRIVIAL(info) << "Output File : " << output->get_filename();
+            BOOST_LOG_TRIVIAL(info) << "Report written to : " << output->get_filename();
             output->write_message(report_result);
         }
+        BOOST_LOG_TRIVIAL(info) << report_result;
         return !is_valid; // A valid file returns an exit code 0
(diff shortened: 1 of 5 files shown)


src/validator_main.cpp
```diff
@@ -189,6 +189,7 @@ int main(int argc, char** argv)
         std::string report_result = "According to the VCF specification, the input file is " + std::string(is_valid ? "" : "not ") + "valid";
         BOOST_LOG_TRIVIAL(info) << report_result;
         for (auto & output : outputs) {
+            BOOST_LOG_TRIVIAL(info) << "Output File : " << output->get_filename();
```
jmmut: the message is a bit more clear as `"Report written to: "`.

srbcheema1:
@jmmut I have made the requested changes.
Please have a look at it.

srbcheema1:
@jmmut I think there may be possibility that some error or exception is generated in try block before writing it to file. so thats why i was preffering not to print it in line 143. As it would be pointless to print it there if no file is written. tell me if I got something wrong.

jmmut:
oh, my fault, the line `std::string report_result = "...` needs to be before the loop

jmmut:
regarding the log placement, we are discussing it, we'll tell you later

jmmut:
ok, after thinking about this, just move the line `std::string report_result = "...` line before the loop to make it compile and we'll merge this. In the future, if we have another report that doesn't write to a file we'll change the name of the method to something more generic, because it will be still useful.

srbcheema1:
@jmmut , should I undo the changes I did in last commit regarding these two lines(moving both the lines together). Or should I move only-begotten one line above the loop.

jmmut:
only one line, we still want the last line to be the final yes/no statement.

Issue #92, opened 2017-10-04
Not all warnings are tested


Anishka0107:
Add tests for warnings

jmmut:
@srbcheema1, are you going to add more tests? As there are several students contributing, it would be good to know if someone is working on an issue, so that no effort is wasted, and this issue has still room for more improvements.

srbcheema1:
I am still working on it.
I was just slow due to some academic work pressure in past few days. I will be soon updating the PR

jmmut:
no hurries! just asking

srbcheema1:
I was stuck somewhere on reading compressed file issue. I sent you the problem and its available solution. please have a look.

tcezard:
Tracked in EVA-2053: closing

PR #119, opened 2018-03-05
Added test cases for svlen

srbcheema1:
This PR aims to solve issue #92 partially

### DIFF
--- a/test/vcf/optional_policy_test.cpp
+++ b/test/vcf/optional_policy_test.cpp
@@ -135,6 +135,51 @@ namespace ebi
                                 vcf::InfoBodyError*);
             }
         }
+
+        SECTION("SVLEN test")
+        {
+                source->meta_entries.emplace(vcf::INFO,
+                    vcf::MetaEntry{
+                        1,
+                        vcf::INFO,
+                        {
+                            { vcf::ID, vcf::SVLEN },
+                            { vcf::NUMBER, "2" },
+                            { vcf::TYPE, vcf::INTEGER },
+                            { vcf::DESCRIPTION, "Difference in length between REF and ALT alleles" }
+                        },
+                        source
+                });
+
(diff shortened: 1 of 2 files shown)


test/vcf/optional_policy_test.cpp
```diff
@@ -135,6 +135,51 @@ namespace ebi
...
+                        source
+                });
+
+                CHECK_NOTHROW( (optional_policy.optional_check_body_entry(parsing_state, vcf::Record{
+                                    1,
+                                    "chr1",
+                                    123456,
+                                    { "id123" },
+                                    "A",
+                                    { "AC","AT" },
```
Anishka0107: Change this to `{ "AC", "ACT" }`.
SVLEN is the difference between the length of ALT and REF columns (as you can see in the Description you added in line 149). To ensure that this does not throw, you should change the values.
You could change other things as well, like the REF column or SVLEN.

Anishka0107:
Though unrelated, but you could also change the line https://github.com/EBIvariation/vcf-validator/blob/develop/src/vcf/validate_optional_policy.cpp#L178 by removing the trailing extra space in the error message "INFO SVLEN should have same number of values as ALT ".

Make it "INFO SVLEN should have same number of values as ALT".

test/vcf/optional_policy_test.cpp
```diff
@@ -135,6 +135,51 @@ namespace ebi
                                 vcf::InfoBodyError*);
             }
         }
+
+        SECTION("SVLEN test")
+        {
+                source->meta_entries.emplace(vcf::INFO,
```
Anishka0107: There seems to be a silly issue with the indentation. I think you used 2 tabs instead of 1 (after the beginning of the SECTION block) in line 41 and below (till 181). Make it 1 tab.

PR #121, opened 2018-03-12
Tests for warnings in the body section

srbcheema1:
This PR aims to solve issue #92 partially. 
Added tests for position, id, reference alternate matching and info imprecise.

### DIFF
--- a/test/vcf/optional_policy_test.cpp
+++ b/test/vcf/optional_policy_test.cpp
@@ -22,7 +22,7 @@
 namespace ebi
 {
 
-    TEST_CASE("Record Position warnings", "[body position warnings]")
+    TEST_CASE("Record Position warnings", "[warnings]")
     {
         std::shared_ptr<vcf::Source> source{
             new vcf::Source{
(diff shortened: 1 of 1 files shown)


srbcheema1:
Thank you for suggesting the changes. I have added them in next commit. Can you please take a look?

test/vcf/optional_policy_test.cpp
```diff
@@ -21,7 +21,229 @@
...
+                                { vcf::PASS },
+                                { { "XYZ", "1" } },
+                                { },
+                                { },
+                                source})),
+                            vcf::IdBodyError*);
+        }
+    }
+
+    TEST_CASE("Record Reference Alternate matching warnings", "[body reference alternate matching warnings]")
```
jmmut: tags are a way to group tests, so it's not really useful to have so many specific tags.
https://github.com/catchorg/Catch2/blob/master/docs/test-cases-and-sections.md#tags
I would keep just `[warnings]` for these tests

srbcheema1: changed :)

PR #123, opened 2018-03-14
Adding Clang 4.0 and 5.0

ManeshNarayan:
Clang version 4.0 and 5.0 added to .travis.yml file

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,53 +8,18 @@ cache:
 
 matrix:
   include:
-    # Clang 3.5
-    - compiler: clang
-      env: C_COMPILER=clang-3.5 CXX_COMPILER=clang++-3.5
-      addons:
-        apt:
-          sources:
-            - boost-latest
-            - ubuntu-toolchain-r-test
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -62,6 +62,42 @@ matrix:
...
 
+    # Clang 4.0
+    - compiler: clang
+      env: C_COMPILER=clang-4.0 CXX_COMPILER=clang++-4.0
+      addons:
+        apt:
+          sources:
+            - boost-latest
+            - ubuntu-toolchain-r-test
+            - llvm-toolchain-trusty-4.0
```
cyenyxe: Given that Trusty is the default Travis environment now, could you please also replace the "llvm-toolchain-precise" packages for Trusty ones?

Anishka0107:
How about removing the older versions of clang and keeping the latest 3.x one? As discussed in PR #110

ManeshNarayan:
The precise packages have been replaced with the trusty packages and the latest clang 3.x (3.9) version is added (as discussed in PR #110 )

Issue #122, opened 2018-03-13
EVA-1016 Flag symbolic SV alleles that are reported as duplicates as Warnings


jmmut:
symbolic alleles (for Structural Variants) are those where the ALT column has a angle bracket enclosed string such as "<INS>", defined in 1.4.5 of the spec https://samtools.github.io/hts-specs/VCFv4.3.pdf

Currently we raise an error if the following two lines appear:
```
1	100	.	A	<INS>	.	.	.
1	100	.	A	<INS>	.	.	.
```
This is not strictly a duplicate, as we don't know if the inserted allele was the same. Therefore we have to put a special case in the code that checks for duplicates in record_cache.hpp: if the alternate allele is enclosed in "<>", don't include it in the cache.

PR #124, opened 2018-03-15
Added check to ignore symbolic alternate alleles

ManeshNarayan:
Updated record_cache.hpp to check for and prevent the addition of symbolic alternate alleles to cache, as mentioned in issue #122.


inc/vcf/record_cache.hpp
```diff
@@ -113,8 +140,26 @@ namespace ebi
             }
         }
 
+        /**
+         *Checks if the alternate allele is symbolic 
+         */ 
+        bool is_symbolic_allele(const RecordCore &record_core)
```
cyenyxe: If you modify the `RecordCore` class to have the variant type as an attribute, this method would be no longer needed.

inc/vcf/record_cache.hpp
```diff
@@ -89,17 +94,39 @@ namespace ebi
                     duplicates.emplace_back(new DuplicationError{record_core.line, ss.str(), duplicate_variant_lines});
                 }
 
-                cache.insert(range.second, record_core);
+                // Insert the record in cache if the allele isn't a symbolic variant
+                if( is_symbolic_allele(record_core) == symbolic_allele_check) {
```
jmmut: we found this a bit confusing. With the current design, each record is checked against the regular cache and the symbolic cache, but only inserted in the corresponding cache. My suggestion is 

1) in the ragel part we aim for this: 
```
cache.check_duplicates(*record);

for r in cache.get_duplicates
    handle_error(r)

for r in cache.get_symbolic_duplicates
    handle_warning(r)
```

2) adding two fields `duplicates` and `symbolic_duplicates`, 
3) the function `check_duplicates` checks `.equal_range(record_core);` against one cache or the other depending on whether the recordCore is symbolic or not.
4) using the same distiction, emplace duplicates in `duplicates` or `symbolic_duplicates`.
4) the function `check_duplicates` doesn't return anything.
5) add two getters: `get_duplicates()` and `get_symbolic_duplicates()`

jmmut:
Also, did you try to compile the project locally? it's fine to do small changes (documentation, comments) directly in github, but that's not the general workflow we use. How familiar with git are you? do you know how to use branch, commit, pull, push and rebase?

ManeshNarayan:
I have tried to integrate all the requested changes as suggested. 
I normally compile and run the program in the docker environment on my machine. I am familiar with git and the mentioned concepts. I will start actively using them.

ManeshNarayan:
The run is failing but the error output on travis-ci shows lines from the upstream(older) vcf.ragel file instead of the vcf.ragel file which is part of the commit. What could be the issue here? It runs fine on my system

jmmut:
I think the issue is that you didn't include in the commits the code that ragel generates.

As we explain in the [readme](https://github.com/EBIvariation/vcf-validator/blob/master/README.md#generate-code-from-descriptors) the code generated by ragel (vcf-validator/inc/vcf/validator_detail_v41.hpp, v42 and v43) should be updated in the repository, so that it's not needed to install ragel to build the project.

However, once we need to modify the ragel code (vcf.ragel) we have to regenerate them with the lines:
```
ragel -G2 src/vcf/vcf_v41.ragel -o inc/vcf/validator_detail_v41.hpp
ragel -G2 src/vcf/vcf_v42.ragel -o inc/vcf/validator_detail_v42.hpp
ragel -G2 src/vcf/vcf_v43.ragel -o inc/vcf/validator_detail_v43.hpp
```

jmmut:
well, now I remembered you used the docker compilation, and that should run the ragel commands for you. Then, it's just a matter of commiting the files validator_detail_v4*.hpp

ManeshNarayan:
I added the validator_detail_v4*.hpp files. However the run is now failing with "No such file or directory" when test_validator is run. Building both the master branch from EBIvariation and the branch from my fork are giving the same error.

jmmut:
I would say the issue you see is that you are running the command from the `bin` directory. in the [readme](https://github.com/EBIvariation/vcf-validator/blob/master/README.md#tests) there's a note explaining that the files that are used in the tests have a path from the root of the project. If you run the tests as `./bin/test_validator` those should go.

however, the tests in travis are actually failing because of a segmentation fault. Let me have a look at those.

ManeshNarayan:
Running it with ./bin/test_validator as mentioned in the README caused the failures. However I tried running it with ./build/bin/test_validator and it worked!

jmmut:
hmm, that's weird, can you post what `git status` says? if travis failed and in my machine fails (with a segfault) it's likely that you didn't commit all changes

ManeshNarayan:
Oh no, this was in reference to the previous comment for running the command for the master branch from EBIvariation. The test_validator for this branch runs correctly when executed with ./build/bin/test_validator.

The commit is throwing a segfault, as you pointed out. I should have been clearer. Apologies.

inc/vcf/normalizer.hpp
```diff
@@ -32,12 +32,22 @@ namespace ebi
...
         
         RecordCore(size_t line, const std::string &chromosome, size_t position, 
                    const std::string &reference_allele, const std::string &alternate_alleles) 
             : line(line), chromosome(chromosome), position(position), 
               reference_allele(reference_allele), alternate_allele(alternate_alleles)
-        { }
+        {
+            // checks the alternate allele to see if it is a symbolic allele.
+            // For this, enclosure by angular brackets is the necessary and sufficient condition
+            if ( (alternate_alleles.find("<") == 0) && (alternate_alleles.find(">") == alternate_alleles.length() - 1) ) {
```
jmmut: what we meant is using the property from Record as another parameter to this constructor

inc/vcf/record_cache.hpp
```diff
@@ -114,7 +158,10 @@ namespace ebi
         }
 
       private:
-        std::multiset<RecordCore> cache;
+        std::multiset<RecordCore> cache_duplicates;
+        std::multiset<RecordCore> cache_symbolic_duplicates;
+        std::vector<Error*> list_duplicates{};
+        std::vector<Error*> list_symbolic_duplicates{};
```
jmmut: not sure what is the reason for the segfault yet, but if you revert this back to std::vector<std::unique_ptr<Error>> it works

inc/vcf/normalizer.hpp
Anishka0107: Hmm yes, it's already implemented in the Record.

And also if the initial idea (no longer required) was to see if the string `alternate_allele` is enclosed by `<>`, this line wouldn't be ideal. Because the find always returns the first occurrence of a character, and in this case you might get `>` before itself (not sure though, maybe the validator flags it wrong before reaching this block of code?) and not at the last index.

Moreover, you can simply check it by seeing if the character at the first index is `<` and the one at the last one is `>`, I think😅. Why go for O(n) when you can do it in O(1)?

ManeshNarayan: Since the tags for all possible structural variants were enclosed within <> and were devoid of angle brackets inside the tag itself, I figured if > is present inside the tag that should be acknowledged as an error.
But I noticed that the Record class uses only the starting angle bracket to identify it as a structural variant, so I guess such errors won't exist at this block of code. And in such a situation, your point about the time complexity is correct.

ManeshNarayan:
The requested changes have been addressed. 
The cases in normalize_test.cpp don't contain structural variants, so the values for normalized_alternate_type haven't been entered for these, as they won't be utilized. That is fine right?

inc/vcf/record_cache.hpp
```diff
@@ -49,8 +49,46 @@ namespace ebi
...
+         * Even if more than one duplicate is found for the same variant during the file validation, each occurrence 
+         * is reported only once:
+         * - If just one processed record is equivalent to the parameter, both are reported in 2 different errors.
+         * - If more than one processed record is equivalent to the new one, only the new one is reported.
+         * 
+         * Nonetheless, if the capacity is too small, it may cause incorrect reporting, such as reporting several times
+         * the first occurrence or failing to report duplicates that are farther apart than the capacity.
+         */
+        std::vector<std::unique_ptr<Error>> get_duplicates()
+        {
```
jmmut: don't put spaces at the end of the line

inc/vcf/record_cache.hpp
```diff
@@ -83,23 +130,25 @@ namespace ebi
 
                     if (++range.first == range.second) {
                         // if only one match, return an extra error for the first occurrence
-                        duplicates.emplace_back(new DuplicationError{first_occurence_line, ss.str()});
+                        duplicates.emplace_back(new DuplicationError{first_occurence_line, ss.str()});
```
jmmut: don't put spaces at the end of the line

inc/vcf/record_cache.hpp
```diff
@@ -83,23 +130,25 @@ namespace ebi
 
                     if (++range.first == range.second) {
                         // if only one match, return an extra error for the first occurrence
-                        duplicates.emplace_back(new DuplicationError{first_occurence_line, ss.str()});
+                        duplicates.emplace_back(new DuplicationError{first_occurence_line, ss.str()});        
                     }
 
                     duplicates.emplace_back(new DuplicationError{record_core.line, ss.str(), duplicate_variant_lines});
+
```
jmmut: unneeded line

inc/vcf/record_cache.hpp
```diff
@@ -83,23 +130,25 @@ namespace ebi
...
                 }
 
-                cache.insert(range.second, record_core);
+                cache.insert(range.second, record_core);        
+
```
jmmut: unneeded line

ManeshNarayan:
Added test cases and made the suggested changes in this commit

test/vcf/record_cache_test.cpp
```diff
@@ -126,5 +139,11 @@ namespace ebi
             // the first occurrence should not be reported again
             CHECK( count_duplicates_and_rethrow_error(cache, {100, "GA", {"GT"}}) == 1 );
         }
+        SECTION("Symbolic Duplicates")
+        {
+            CHECK( count_duplicates_and_rethrow_error(cache, {107, "C", {"<INS>"}}) == 0 );
+            // This won't throw an error. Only warning would be generated, which isn't detected here.
+            CHECK( count_duplicates_and_rethrow_error(cache, {107, "C", {"<INS>"}}) == 0 );
```
jmmut: These tests are incomplete because the function `count_duplicates_and_rethrow_error` is using only `get_duplicates` and not `get_symbolic_duplicates`. The comment above is misleading because the code flow does not involve warnings.

Another `TEST_CASE` should be added that uses `get_symbolic_duplicates` and test that with both symbolic and regular variants.

test/vcf/record_cache_test.cpp
```diff
@@ -142,8 +153,43 @@ namespace ebi
         SECTION("Symbolic Duplicates")
         {
             CHECK( count_duplicates_and_rethrow_error(cache, {107, "C", {"<INS>"}}) == 0 );
-            // This won't throw an error. Only warning would be generated, which isn't detected here.
+            // This won't throw an error.
```
jmmut: This comments don't add information. `count_duplicates_and_rethrow_error` should only throw in case of inner malfunction. if there are no duplicates it returns 0, and if there are duplicates it returns the count. maybe the function should be called `count_duplicates_or_rethrow_error`, or `count_duplicates` to hide implementation details.

test/vcf/record_cache_test.cpp
```diff
@@ -142,8 +153,43 @@ namespace ebi
...
+    {
+        vcf::RecordCache cache{0};
+
+        cache.check_duplicates(build_mock_record({100, "A", {"T"}}));
+        cache.check_duplicates(build_mock_record({101, "A", {"T"}}));
+        cache.check_duplicates(build_mock_record({102, "A", {"T"}}));
+        cache.check_duplicates(build_mock_record({103, "A", {"T"}}));
+        cache.check_duplicates(build_mock_record({104, "A", {"T"}}));
+        cache.check_duplicates(build_mock_record({105, "A", {"T"}}));
+        cache.check_duplicates(build_mock_record({107, "C", {"<INS>"}}));
```
jmmut: half of these are not relevant. keep only those that allow interesting test checks

test/vcf/record_cache_test.cpp
```diff
@@ -142,8 +153,43 @@ namespace ebi
         SECTION("Symbolic Duplicates")
         {
             CHECK( count_duplicates_and_rethrow_error(cache, {107, "C", {"<INS>"}}) == 0 );
-            // This won't throw an error. Only warning would be generated, which isn't detected here.
+            // This won't throw an error.
             CHECK( count_duplicates_and_rethrow_error(cache, {107, "C", {"<INS>"}}) == 0 );
         }
     }
+
+    TEST_CASE("Symbolic Duplicates test")
```
jmmut: keep the style of the other messages: "RecordCache tests: symbolic duplicates

test/vcf/test_utils.hpp
```diff
@@ -38,6 +38,7 @@ namespace ebi
       size_t normalized_pos;
       std::string normalized_reference;
       std::string normalized_alternate;
+      vcf::RecordType normalized_alternate_type;
```
jmmut: is this line (and the change in normalize_test.cpp) necessary? the type from TestRecord is not used anywhere I can see.

src/vcf/vcf.ragel
```diff
@@ -219,12 +219,21 @@
         try {
             // Handle all columns and build record
             ParsePolicy::handle_body_line(*this);
+
```
jmmut: it's usually ugly to have 2 or more empty lines, especially within functions

test/vcf/test_utils.hpp
ManeshNarayan: The change in normalize_test.cpp is needed because the constructor for RecordCore class has been changed to use the value directly from the Record class. The change in the struct TestRecord was done so that the change in normalize_test.cpp could be accomodated. 
It should be possible to pass an empty RecordType object in normalize_test.cpp and not change RecordType, but I thought it would lose consistency.

jmmut: mmm, some change was indeed needed in normalize_test.cpp, given that it creates a RecordCore, but TestRecord is not used for anything using the type. TestRecord is meant to lack lots of fields in Record, it just has the fields needed in the normalize_test.cpp. 

Even with your change, `normalized_alternate_type` will always be empty (implicitly), so you can just hardcode an empty string in `test_normalization` as we are hardcoding other fields. Those are clearly mocks and not intended to work outside their narrow tests.

test/vcf/record_cache_test.cpp
```diff
@@ -107,24 +131,60 @@ namespace ebi
...
+    TEST_CASE("RecordCache tests: symbolic duplicates")
+    {
+        vcf::RecordCache cache{0};
+
+        cache.check_duplicates(build_mock_record({100, "A", {"T"}}));
+        cache.check_duplicates(build_mock_record({101, "A", {"T"}}));
+        cache.check_duplicates(build_mock_record({103, "A", {"T"}}));
+        cache.check_duplicates(build_mock_record({107, "C", {"<INS>"}}));
+
+        SECTION("No symbolic duplicates") {
```
cyenyxe: `{` should be in new line

test/input_files/v4.3/passed/passed_symbolic_duplicates.vcf
```diff
@@ -0,0 +1,4 @@
+##fileformat=VCFv4.3
```
cyenyxe: Do we need 4.1 and 4.2 test files?

inc/vcf/record_cache.hpp
```diff
@@ -60,15 +78,44 @@ namespace ebi
...
-            std::vector<std::unique_ptr<Error>> duplicates{};
 
+            list_duplicates.clear();
+            list_symbolic_duplicates.clear();
+            
             for (RecordCore &record_core: record_cores) {
-                std::pair<std::multiset<RecordCore>::iterator, std::multiset<RecordCore>::iterator> range =
-                        cache.equal_range(record_core);
 
+                // create references to the appropriate data structures for th alternate allele type
```
cyenyxe: I would be happy using `auto` in the next 3 statements, given they are assigning from attributes whose type we already now.

test/input_files/v4.3/passed/passed_symbolic_duplicates.vcf
ManeshNarayan: I don't think they are needed because there would be no change in the input files except for the version number as the specs don't diverge for this particular example. However, I noticed that there are other inputs where this would be true as well but they were added. So for the sake of completeness I am adding test files for 4.1 and 4.2.

PR #130, opened 2018-03-27
validator: Allow AF Info field to contain missing value.

blankenberg:
Also, Fix for text report AF error.

### DIFF
--- a/src/vcf/record.cpp
+++ b/src/vcf/record.cpp
@@ -377,7 +377,7 @@ namespace ebi
             for (auto & value : values) {
                 if (value == MISSING_VALUE) { continue; }
                 if (std::stold(value) < 0 || std::stold(value) > 1) {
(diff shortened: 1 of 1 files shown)


jmmut:
Good spot! thanks for the contribution.

The code you added to allow missing values not only allows simple missing values, e.g. `AF=.` but also to mixed (missing and non-missing) values such as `AF=.,0.3`, which we need to check with the spec to see if it's valid before merging.

Also, this made us realise we need to allow missing in other fields as well.

blankenberg:
Great, let me know what you are able to decipher as far as the spec is concerned.

I'll also just briefly mention that the original error was quite cryptic (in contrast to other 'real' validation errors) as it was just an uncaught stold conversion. It may be worthwhile to check for additional places in the code and where this sort of unchecked error could occur and provide one of the much more meaningful errors that vcf_validator generally provides.
```
./vcf_validator -i ../../test.vcf 
[error] stold
```

PR #139, opened 2018-05-31
Code and test cases for assembly checker

ManeshNarayan:


test/vcf/assembly_checker_test.cpp
```diff
@@ -0,0 +1,135 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "assembly_checker_test_aux.hpp"
+
+namespace ebi
+{	
+	
+	TEST_CASE("Valid fasta and vcf combination", "[assembly_checker]")
```
jmmut: these are not really unitary tests. extract a function from the main loop to be able to test just one VcfVariant here

src/vcf/assembly_checker.cpp
```diff
@@ -0,0 +1,101 @@
...
+
+          std::set<std::string> absent_chromosomes;
+          size_t num_matches = 0;
+          size_t num_variants = 0;
+
+          // Reading FASTA index, and querying FASTA file
+          auto index = bioio::read_fasta_index(fasta_index_input);
+
+          // Reading VCF file
+          while (util::readline(vcf_input, vector_line).size() != 0) {
```
jmmut: try to think and extract functions from inside the loop. experiment with ideas like:
- function for checks on preconditions
- function for checking a VcfVariant
- function for keeping track of counts

try a structure where each function can be described (and named) as doing just one thing

test/vcf/assembly_checker_test_aux.hpp
```diff
@@ -0,0 +1,38 @@
...
+#include "vcf/string_constants.hpp"
+
+namespace ebi
+{
+  	inline std::string get_file_prefix(std::string folder)
+  	{
+  		std::vector<std::string> path_components;
+  		boost::split(path_components, folder, boost::is_any_of("/"));
+  		return path_components[path_components.size() - 2];
+  	}
```
cyenyxe: Avoid dealing with paths as string whenever possible, because separators are OS-specific (and one of the goals of the other GSoC project is to make this work on MacOS and Windows).

You can use [this code from the validator test suite](https://github.com/EBIvariation/vcf-validator/blob/master/test/vcf/parser_v43_test.cpp#L52-L65) as reference.

test/vcf/assembly_checker_integration_test.cpp
```diff
@@ -0,0 +1,93 @@
...
+      }
+  }
+
+  TEST_CASE("Invalid VCF and Fasta combination", "[assembly_checker]")
+  {
+      SECTION("Single entry, single mismatch")
+      {
+          std::string folder = "test/input_files/assembly_checker/failed/failed_singlemismatch_singleentry/";
+          std::string file_prefix = get_file_prefix(folder);
+          CHECK_FALSE(ebi::vcf::assembly_checker::check_vcf_ref(folder + file_prefix + ebi::vcf::VCF_EXT, folder + file_prefix + ebi::vcf::FASTA_EXT, folder + file_prefix + ebi::vcf::INDEX_EXT));
```
cyenyxe: I like this test suite a lot more now :) You can check [this](https://github.com/EBIvariation/vcf-validator/blob/master/test/vcf/parser_v43_test.cpp#L52-L65) and [this](https://github.com/EBIvariation/vcf-validator/blob/master/test/vcf/parser_test_aux.hpp#L34-L41) on how to test the entry point method using streams instead of file paths as arguments.

Note: In fact one of the methods linked receives a file path but that could be stdin.

CMakeLists.txt
```diff
@@ -179,6 +188,10 @@ target_link_libraries (test_validator ${LIBRARIES_TO_LINK})
 enable_testing ()
 add_test (NAME ValidatorTests COMMAND test_validator)
 
+add_executable (test_assembly_checker test/main_test.cpp ${ASSEMBLY_CHECKER_TESTS})
+target_link_libraries (test_assembly_checker ${LIBRARIES_TO_LINK})
+enable_testing ()
+add_test (NAME AssemblyCheckerTests COMMAND test_assembly_checker)
```
cyenyxe: Now that we have 3 tools we should agree how to structure the test binaries. `ALL_TESTS` includes validator and debugulator tests, but the assembly checker ones are in a completely separate set. I would say that "ALL" should mean "ALL", but it is also true that we should make the granularity more clear and try to remove redundancies.

Any particular preference @jmmut ?

src/assembly_checker_main.cpp
```diff
@@ -0,0 +1,104 @@
...
+
+    auto vcf_path = vm[ebi::vcf::VCF].as<std::string>();
+    auto fasta_path = vm[ebi::vcf::FASTA].as<std::string>();
+    boost::filesystem::path fasta_boost_path{fasta_path};
+    auto fasta_index_path = fasta_path + ".fai";
+
+    try {
+
+        if (!ebi::vcf::assembly_checker::check_vcf_ref(vcf_path, fasta_path, fasta_index_path)) {
+            BOOST_LOG_TRIVIAL(info) << "VCF and reference fasta are not matching";
```
cyenyxe: FASTA insted of fasta, please check in other places too.

src/assembly_checker_main.cpp
```diff
@@ -0,0 +1,104 @@
...
+        }
+
+        return 0;
+
+    } catch (std::exception const &ex) {
+        BOOST_LOG_TRIVIAL(error) << "Aborting execution, error: " << ex.what();
+        return 1;
```
cyenyxe: [Exception handling from the validator](https://github.com/EBIvariation/vcf-validator/blob/master/src/validator_main.cpp#L197-L205) to be used as reference.

src/vcf/assembly_checker.cpp
```diff
@@ -0,0 +1,123 @@
...
+
+              auto fasta_sequence = bioio::read_fasta_contig(fasta_input, index.at(vcf_variant.chromosome),
+                                                       vcf_variant.position - 1, vcf_variant.reference_allele.length());
+              auto reference_sequence = vcf_variant.reference_allele;
+
+              match_stats.add_match_result(is_matching_sequence(fasta_sequence, reference_sequence));
+          }
+
+          BOOST_LOG_TRIVIAL(info) << "Number of matches: " << match_stats.num_matches << "/" << match_stats.num_variants;
+          BOOST_LOG_TRIVIAL(info) << "Percentage of matches: " << (static_cast<double>(match_stats.num_matches) / match_stats.num_variants) * 100 << "%";
```
cyenyxe: This will break if the VCF file contains no variants, which is a valid (although pretty much useless) case.

src/vcf/assembly_checker.cpp
```diff
@@ -0,0 +1,123 @@
...
+                  continue;
+              }
+
+              if (vcf_variant.position == 0) {
+                  BOOST_LOG_TRIVIAL(warning) << "Position 0 should only be used for a telomere";
+                  continue;
+              }
+
+              auto fasta_sequence = bioio::read_fasta_contig(fasta_input, index.at(vcf_variant.chromosome),
+                                                       vcf_variant.position - 1, vcf_variant.reference_allele.length());
```
cyenyxe: Fix indentation, so the parameters declaration is vertically aligned.

src/vcf/assembly_checker.cpp
```diff
@@ -0,0 +1,123 @@
...
+              auto reference_sequence = vcf_variant.reference_allele;
+
+              match_stats.add_match_result(is_matching_sequence(fasta_sequence, reference_sequence));
+          }
+
+          BOOST_LOG_TRIVIAL(info) << "Number of matches: " << match_stats.num_matches << "/" << match_stats.num_variants;
```
cyenyxe: Instead of writing the report here do it from the main method. That will make this one be focused just on checking that the reference is correct, not doing UI-related tasks.

You can check the following existing code:
* [This method](https://github.com/EBIvariation/vcf-validator/blob/master/src/validator_main.cpp#L112-L150) for how to select a report type. We will only have the `MatchStats` for now, somewhat close to the error summary report in the validator. An ODB report will be added later on.
* [Entry point method](https://github.com/EBIvariation/vcf-validator/blob/master/src/vcf/validator.cpp#L133) for the validation. It can give you ideas on how to split pre-validation checks and also how to write the report. Try to follow this structure as closely as possible to improve consistency of the code across tools.
* [This class](https://github.com/EBIvariation/vcf-validator/blob/master/inc/vcf/summary_report_writer.hpp) containing the `SummaryReportWriter` mentioned above.

CMakeLists.txt
jmmut: I'm used to just run the `test_validator` binary which runs all tests. I'm not sure having more binaries adds value, given that we could use the tag system better to select a subset of tests.

cyenyxe: Good point about the tag system. Creating a single binary for all the tests should reduce the build time. @ManeshNarayan can you try replacing the current test command as suggested by @jmmut ? At least to split the validator and assembly tests in 2 runs for the sake of traceability.

src/vcf/assembly_checker.cpp
cyenyxe: NaN is an acceptable output. I thought it would just crash, because that is what happens in other languages.

CMakeLists.txt
cyenyxe: +1 at using ALL_TESTS. About the split, I meant using the tags to execute first only the validator tests, then the assembly tests for improved granularity. But it is not critical so I'm fine with running all of them in a single run.

src/vcf/assembly_report_writer.cpp
```diff
@@ -0,0 +1,134 @@
...
+            write_match(vcf_variant);
+        } else {
+            write_mismatch(vcf_variant);
+        }
+    }
+
+	OdbAssemblyReportWriter::OdbAssemblyReportWriter(const std::string &db_name) : db_name(db_name), current_transaction_size{0},
+                                                          transaction_size{1000000}
+	{	
+		try {
```
jmmut: Even though this is a work in progress, remember to use spaces instead of tabs. I think it's faster and easier to do that if you configure your editor to use spaces, and tabs don't even appear.

inc/vcf/assembly_report_writer.hpp
```diff
@@ -0,0 +1,81 @@
...
+  namespace vcf
+  {   
+
+    class AssemblyReportWriter
+    {
+      public:
+        virtual void write_results() = 0;
+        virtual void write_mismatch(const vcf::VcfVariant &vcf_variant) = 0;
+        virtual void write_match(const vcf::VcfVariant &vcf_variant) = 0;
+        virtual void add_result(bool result, const vcf::VcfVariant &vcf_variant) = 0;
```
jmmut: if you prefer to use the AssemblyReportWriter interface using this method, keep the others (write_mismatch and  write_match) private, not visible in the interface declaration

inc/vcf/assembly_report_writer.hpp
```diff
@@ -0,0 +1,81 @@
...
+        virtual void write_match(const vcf::VcfVariant &vcf_variant) override;
+        virtual void write_results() override;
+        virtual void add_result(bool result, const vcf::VcfVariant &vcf_variant) override;
+
+      private:
+    	MatchStats match_stats;
+
+    };
+
+    class OdbAssemblyReportWriter : public AssemblyReportWriter
```
jmmut: agreed to make OdbAssemblyReportWriter implement also `AssemblyReportReader` to make testing easier. we can do a first version of this new interface with just a method like `MatchStats read_result()`.

tcezard:
This has been implemented elsewhere

PR #141, opened 2018-06-12
[W.I.P] Simplyfying build for windows

srbcheema1:
This PR aims to simplify build for windows operating system.


CMakeLists.txt
```diff
@@ -47,22 +50,28 @@ if (OSX)
...
-set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -Wall -Wno-unknown-pragmas")
-# Needed to statically link ODB
-set (CMAKE_C_FLAGS " -DSQLITE_ENABLE_UNLOCK_NOTIFY -DSQLITE_OMIT_LOAD_EXTENSION")
-set (CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/bin)
-if (NOT CMAKE_BUILD_TYPE)
-    set (CMAKE_BUILD_TYPE  "RelWithDebInfo" CACHE STRING "Choose the type of build, options are: None(CMAKE_CXX_FLAGS or CMAKE_C_FLAGS used) Debug Release RelWithDebInfo MinSizeRel." FORCE)
+if (WINDOWS)
+    include_directories (lib/sqlite)# windows specific
+    include_directories (inc/windows_specific)# windows specific
+    set (CMAKE_BUILD_TYPE  "Release" CACHE STRING "On windows we only support release mode" FORCE)
```
cyenyxe: Then what about the plan we suggested for Travis to run in "Debug" mode to make it faster?

CMakeLists.txt
```diff
@@ -202,6 +227,8 @@ set (ALL_TESTS
         test/vcf/test_utils.hpp
         )
 
+set (CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/bin)
```
cyenyxe: Were we forcing this before? I don't think we should.

CMakeLists.txt
```diff
@@ -88,22 +97,29 @@ elseif (OSX)
...
-    )
+    #add_custom_command (
+            #OUTPUT ${CMAKE_HOME_DIRECTORY}/${right_place_gencppfile}
+            #COMMAND cd ${CMAKE_HOME_DIRECTORY}/ && odb --include-prefix vcf --std c++11 -d sqlite --generate-query --generate-schema
+            #--hxx-suffix .hpp --ixx-suffix .ipp --cxx-suffix .cpp
+            #--output-dir ${CMAKE_HOME_DIRECTORY}/inc/vcf/ ${_file}
+            #&& mv ${CMAKE_HOME_DIRECTORY}/${gencppfile} ${CMAKE_HOME_DIRECTORY}/${right_place_gencppfile}
+            #DEPENDS ${_file}
+            #COMMENT "Building odb for ${_file}"
+    #)
```
cyenyxe: This should be uncommented.

CMakeLists.txt
```diff
@@ -47,22 +50,28 @@ if (OSX)
...
 
 
-# no unknown pragmas: ODB compiler uses some pragmas that the regular compiler doesn't need
-set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -Wall -Wno-unknown-pragmas")
-# Needed to statically link ODB
-set (CMAKE_C_FLAGS " -DSQLITE_ENABLE_UNLOCK_NOTIFY -DSQLITE_OMIT_LOAD_EXTENSION")
-set (CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/bin)
-if (NOT CMAKE_BUILD_TYPE)
-    set (CMAKE_BUILD_TYPE  "RelWithDebInfo" CACHE STRING "Choose the type of build, options are: None(CMAKE_CXX_FLAGS or CMAKE_C_FLAGS used) Debug Release RelWithDebInfo MinSizeRel." FORCE)
+if (WINDOWS)
```
cyenyxe: This block is a mix of different stuff, for instance:

* Directory inclusion is listed in lines 15-16
* Boost_USE_STATIC_LIBS is enabled for other operative systems in lines 43 and 49, and it is the same for all OS
* CMAKE_C_FLAGS is the same for all OS

So to sum up, this looks fine but need a bit of organization.

test/vcf/debugulator_integration_test.cpp
```diff
@@ -74,7 +74,7 @@ namespace ebi
 
       boost::filesystem::remove(fixed_vcf_path);
 
-      if (not successfully_fixed) {
+      if (! successfully_fixed) {
```
cyenyxe: Remove unnecessary whitespace, it was fine with "not" because it was 2 word, but not with the exclamation mark. It might apply to other files.

CMakeLists.txt
srbcheema1: I will set it to `cache` instead of `force` once I confirm that debug will work fine.
then we can pass the argument `cmake -DCMAKE_BUILD_TYPE=Debug`

CMakeLists.txt
srbcheema1: I put them all in one block, else there would have been a lot of if-else blocks in our cmake. We could compare it with the other version of CMakeLists.txt, then we can discuss which way to keep.

test/vcf/debugulator_integration_test.cpp
srbcheema1: It was only in this file. I will remove it.

srbcheema1:
@cyenyxe @jmmut this PR is ready. window build is working fine. All the tests have passed. Please review :)

CMakeLists.txt
```diff
@@ -145,27 +145,6 @@ add_library (mod_odb STATIC ${MOD_ODB_SOURCES})
 target_link_libraries (mod_odb ${THIRD_PARTY_LIBRARIES})
 
 set (MOD_VCF_SOURCES
-  inc/vcf/debugulator.hpp
```
jmmut: this is not for including the libraries, it's for adding them in the dependency graph, so that when a change is detected in one of these files, all the objects that depend on them are recompiled. Please restore this part.

Issue #142, opened 2018-06-21
Does not fail when output folder does not have write permission


j-coll:
I have executed the validator without specifying the `-o`, so as default uses the location of the input file, where I don't have write permissions. The program has not failed at any point.

```
jcoll@node-01:~$ ./vcf_validator_linux -i /data/my_vcf.vcf.gz
[info] Reading from input file...
[warning] Detected .gz compression
Lines read: 10000
Lines read: 20000
Lines read: 30000
....
....
Lines read: 130610000
Lines read: 130620000
Lines read: 130630000
[info] Summary report written to : /data/my_vcf.vcf.gz.errors_summary.1529575578873.txt
[info] According to the VCF specification, the input file is not valid
jcoll@node-01:~$ less /data/my_vcf.vcf.gz.errors_summary.1529575578873.txt
/data/my_vcf.vcf.gz.errors_summary.1529575578873.txt: No such file or directory
```

jmmut:
This is merged in develop, but we will close when it's in a release and the requestor confirms it's fixed.

jmmut:
I tested this with the latest release and it works as expected.

However, we can improve this message in the case of ODB report:

> ./vcf_validator_linux -i no_write_permissions/a.vcf  -r database
> [error] The validation could not be completed: ODB report: Can't initialize database: 14: unable to open database file

should say the path of the "database file"

PR #143, opened 2018-06-21
break if file is not written

srbcheema1:
this PR closes the issue #142 
now the vcf_validator will throw runtime error is it is unable to write the output file.

```
srb@srb-pc:~$ ./vcf_validator -i /data/my_vcf.vcf.gz
[error] The validation could not be completed: Unable to write output file stdin.errors_summary.1529587453100.txt
```

### DIFF
--- a/inc/vcf/report_writer.hpp
+++ b/inc/vcf/report_writer.hpp
@@ -44,7 +44,9 @@ namespace ebi
             FileReportWriter(std::string filename) : file_name(filename)
             {
                 file.open(filename, std::ios::out);
-                if(!file) throw std::runtime_error{"Unable to write output file " + file_name};
+                if(!file) {
+                    throw std::runtime_error{"Unable to write output file " + file_name};
+                }
             }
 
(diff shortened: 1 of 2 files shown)


inc/vcf/report_writer.hpp
```diff
@@ -44,6 +44,7 @@ namespace ebi
             FileReportWriter(std::string filename) : file_name(filename)
             {
                 file.open(filename, std::ios::out);
+                if(!file) throw std::runtime_error{"Unable to write output file " + file_name};
```
jmmut: Remember we use braces always

jmmut:
still, this is one of fastest resolutions ever of an issue!

jmmut:
The unnoticed error is fixed, but maybe we can improve this message in the case of ODB report:

> ./vcf_validator_linux -i no_write_permissions/a.vcf  -r database
> [error] The validation could not be completed: ODB report: Can't initialize database: 14: unable to open database file

should say the path of the "database file"

PR #146, opened 2018-07-05
Updates supported compressed file and supported platform details

srbcheema1:
This PR is to add details of supported compressed files and supported platform details.
Also some help is added to install_dependencies.bat

### DIFF
--- a/install_dependencies.bat
+++ b/install_dependencies.bat
@@ -1,5 +1,8 @@
 @echo off
 
+if %1'==/?' goto :help
+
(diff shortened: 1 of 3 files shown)


install_dependencies.bat
```diff
@@ -1,5 +1,8 @@
 @echo off
 
+if %1'==/?' goto :help
```
jmmut: let's keep this file simple: revert the changes in this file. then rebase to develop, please

PR #150, opened 2018-07-24
[W.I.P] Assemble Checker Tool

srbcheema1:
This PR targets to implement a simple version of assembly checker. This will include functionality to count the number of matches and percentage of matches. It will also write the problem_lines(lines containing mismatch) to an output file.


srbcheema1:
@cyenyxe @jmmut this PR is ready with code and test cases. Please review the code

src/vcf/assembly_checker.cpp
```diff
@@ -0,0 +1,102 @@
...
+      {
+          std::transform(fasta_sequence.begin(), fasta_sequence.end(), fasta_sequence.begin(), ::tolower);
+          std::transform(reference_sequence.begin(), reference_sequence.end(), reference_sequence.begin(), ::tolower);
+
+          return fasta_sequence == reference_sequence;
+      }
+
+      void check_missing_chromosomes(std::set<std::string> absent_chromosomes)
+      {
+          if (absent_chromosomes.size() > 0) {
```
cyenyxe: This will make the process abort if checking against FASTA files split by chromosome. Instead of checking whether any chromosomes are absent, we should check if **all** are.

As long as one is shared across the VCF and FASTA files, then we can run some validation. We can raise a warning is some are not, because it is true that could help to detect mistakes when running the tool; but it should be a logging message instead of an exception.

src/vcf/vcf_fasta_variant.cpp
```diff
@@ -0,0 +1,51 @@
...
+        position = static_cast<size_t>(std::stoi(record_columns[1]));
+        reference_allele = record_columns[3];
+    }
+
+    std::string VcfVariant::format_chromosome(std::string const &chromosome)
+    {
+        std::string chromosome_copy;
+        std::transform(chromosome.begin(), chromosome.end(), chromosome_copy.begin(), ::tolower);
+        if (boost::starts_with(chromosome_copy, CHR)) {
+            return chromosome.substr(CHR.length());
```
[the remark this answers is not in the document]
srbcheema1: The FASTA files that I analysed in test cases provided by manesh were not containing this Prefix. also FASTA_INDEX file doesn't seems to contain this Prefix. so might be possibility that we may not get the chrom ID in Index file. in the line :

https://github.com/srbcheema1/vcf-validator/blob/8e23e4ab8afbff4a63d01700471fc784be7ea3b2/src/vcf/assembly_checker.cpp#L48

In case when the Chrom id starts with CHR in vcf file where as this is absent in index file and fasta file.

inc/vcf/assembly_report_reader.hpp
```diff
@@ -0,0 +1,40 @@
...
+#define ASSEMBLY_REPORT_READER_HPP
+
+#include <memory>
+#include "vcf/error.hpp"
+
+namespace ebi
+{
+  namespace vcf
+  {
+    class AssemblyReportReader
```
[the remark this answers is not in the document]
srbcheema1: I had just kept it for next PR where it would be useful in odbReportRW. Dropping this for now, is also fine.

src/vcf/vcf_fasta_variant.cpp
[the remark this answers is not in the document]
cyenyxe: But I'm talking about FASTA files in general, not just those in the test datasets. Real VCF and FASTA files we have processed named the same chromosome as:

* 1
* chr1
* chrom1
* chromosome1

We just can't assume a prefix, we need to accept the data as is (and will deal with aliases in a slightly different manner later on).

inc/vcf/vcf_fasta_variant.hpp
```diff
@@ -0,0 +1,43 @@
...
+#define VCF_VCF_FASTA_VARIANT_HPP
+
+#include <string>
+
+namespace ebi
+{
+  namespace vcf
+  {
+
+    class VcfVariant
```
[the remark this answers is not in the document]
srbcheema1: @cyenyxe name `RecordCore` was conflicting with `inc/vcf/normalizer.hpp`.
For now, I have named it Record_Core. I know I have mixed CamelCase and snake_case. We will have to think about some new name for it.

src/vcf/assembly_checker.cpp
```diff
@@ -23,6 +23,19 @@ namespace ebi
     namespace assembly_checker
     {
 
+      Record_Core build_record_core(std::string const & line)
```
srbcheema1: can we use constructor overloading instead ? one for (line) and another for (line, chromosome, position, reference_allele, alternate_allele)

cyenyxe: It is better not to make objects aware of the source they are constructed from, but receive the appropriate parameters.

inc/vcf/normalizer.hpp
```diff
@@ -27,64 +27,94 @@ namespace ebi
...
+                   RecordType alternate_allele_types
+                   ) : line_num(line_num),
+                       chromosome(chromosome),
+                       position(position),
+                       reference_allele(reference_allele),
+                       alternate_allele(alternate_alleles),
+                       alternate_allele_type(alternate_allele_types)
+        {
+            line = "";
+        }
```
cyenyxe: In C++11 you can make one constructor call another one, which makes sure that for a given set of parameters they will behave the same way.

src/vcf/assembly_checker.cpp
```diff
@@ -0,0 +1,113 @@
...
+
+          size_t line_num = 0;
+          while (util::readline(vcf_input, vector_line).size() != 0) {
+              std::string line{vector_line.begin(), vector_line.end()};
+
+              if (boost::starts_with(line, "#")) {
+                  continue;
+              }
+
+              RecordCore record_core = build_record_core(line,++line_num);
```
cyenyxe: `line_num` should not be incremented here because it won't count the header lines. But you could replace the while with a for loop:

```c++
for (size_t line_num = 1; util::readline(vcf_input, vector_line).size() != 0; line_num++) {
    // ...
    RecordCore record_core = build_record_core(line, line_num);
    // ...
}
```

srbcheema1:
@cyenyxe I have made the requested changes. The code is ready to be merged.

PR #151, opened 2018-08-06
Updates install scripts

srbcheema1:
This PR is to solve the issue of bzip and zlib libraries in linux.

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,31 +88,49 @@ The easiest way to build vcf-validator is using the Docker image provided with t
 
 Executables will be created in the `build/bin` subfolder.
 
+
 ### Linux
+
+The build has been tested on the following distros:
+* Ubuntu
+* Arch
 
 The build has been tested on the following compilers:
 * Clang 3.9 to 5.0
 * GCC 4.8 to 6.0
 
+#### Using install_dependencies.sh (Recommended for normal users)
+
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -88,31 +88,49 @@ The easiest way to build vcf-validator is using the Docker image provided with t
 
 Executables will be created in the `build/bin` subfolder.
 
+
 ### Linux
 
+The build has been tested on the following distros:
+* Ubuntu
+* Arch
```
cyenyxe: I'm somewhat reluctant to state this, because in the future we will only test the build **for sure** under Travis' environment.

README.md
```diff
@@ -88,31 +88,49 @@ The easiest way to build vcf-validator is using the Docker image provided with t
...
+
 The build has been tested on the following compilers:
 * Clang 3.9 to 5.0
 * GCC 4.8 to 6.0
 
-#### Dependencies
+#### Using install_dependencies.sh (Recommended for normal users)
```
cyenyxe: Rename this as 'Dependencies', first describing them without the installation instructions, then point to the installation script and drop the instructions for the manual build.

README.md
srbcheema1: hmm sounds reasonable

README.md
```diff
@@ -219,6 +237,24 @@ odb --include-prefix vcf --std c++11 -d sqlite --generate-query --generate-schem
 mv inc/vcf/error-odb.cpp src/vcf/error-odb.cpp
 ```
 
+## Miscellaneous
```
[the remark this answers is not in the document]
srbcheema1: I have dropped the subsection of commands and kept windows section as discussed over the call.

PR #153, opened 2018-08-24
add support to read vcf files from stdin

srbcheema1:
This PR is to add support of reading vcf files directly from stdin in assembly_checker tool.
Now we could also check dynamically generated vcf files through stdin using 

```
vcf_assembly_checker -f /path/to/fasta_file < /path/to/vcf_file
```
**OR** 

```
cat file.vcf | vcf_assembly_checker -f /path/to/fasta_file
```


test/vcf/assembly_checker_integration_test.cpp
```diff
@@ -72,40 +75,44 @@ namespace ebi
         {
             auto folder = boost::filesystem::path("test/input_files/v4.3/assembly_checker/failed/failed_singlemismatch_singleentry/");
             std::string file_prefix = folder.parent_path().filename().string();
-            std::ifstream vcf_input{folder.string() + file_prefix + ebi::vcf::VCF_EXT};
+            std::string vcf_path = folder.string() + file_prefix + ebi::vcf::VCF_EXT;
+            std::ifstream vcf_input{vcf_path};
             std::ifstream fasta_input{folder.string() + file_prefix + ebi::vcf::FASTA_EXT};
             std::ifstream fasta_index_input{folder.string() + file_prefix + ebi::vcf::FASTA_EXT + ebi::vcf::INDEX_EXT};
-            CHECK_FALSE(ebi::vcf::assembly_checker::check_vcf_ref(vcf_input, fasta_input, fasta_index_input, outputs));
+            CHECK(ebi::vcf::assembly_checker::check_vcf_ref(vcf_input, vcf_path, fasta_input, fasta_index_input, outputs));
```
jmmut: Why are these `CHECK` if there are mistmatches?

src/vcf/assembly_checker.cpp
```diff
@@ -74,6 +107,107 @@ namespace ebi
           return outputs[0]->is_valid_report();
       }
 
+      void create_uncompressed_stream(std::istream & input,
```
jmmut: did you copy all this code after deciding it's not worth to extract, or are you working on extracting it?

test/vcf/assembly_checker_integration_test.cpp
srbcheema1: ohh! my mistake. sorry.

test/vcf/assembly_checker_integration_test.cpp
```diff
@@ -32,34 +32,37 @@ namespace ebi
...
+            //auto folder = boost::filesystem::path("test/input_files/v4.3/assembly_checker/passed/passed_empty_vcf/");
+            //std::string file_prefix = folder.parent_path().filename().string();
+            //std::string vcf_path = folder.string() + file_prefix + ebi::vcf::VCF_EXT;
+            //std::ifstream vcf_input{vcf_path};
+            //std::ifstream fasta_input{folder.string() + file_prefix + ebi::vcf::FASTA_EXT};
+            //std::ifstream fasta_index_input{folder.string() + file_prefix + ebi::vcf::FASTA_EXT + ebi::vcf::INDEX_EXT};
+            //CHECK(ebi::vcf::assembly_checker::check_vcf_ref(vcf_input, vcf_path, fasta_input, fasta_index_input, outputs));
+        //}
 
         SECTION("Single entry")
```
jmmut: Now I think of this, there's no new test that reads from a compressed VCF, right? As general rule, whenever you add a new feature, you should tests for it. it improves the maintenance and shows how the feature can be used.

srbcheema1: hmm. i wasto add those tests also. i was thinking to add those tests after i have extracted out the compression functionality. as i will re restructuring few functions tests will also have to be restructured.

jmmut: If you think it's worth to wait for the restructure to do the tests, then include the restructure in this PR

test/vcf/assembly_checker_integration_test.cpp
```diff
@@ -72,40 +86,56 @@ namespace ebi
...
-            std::ifstream vcf_input{folder.string() + file_prefix + ebi::vcf::VCF_EXT};
+            std::string vcf_path = folder.string() + file_prefix + ebi::vcf::VCF_EXT;
+            std::ifstream vcf_input{vcf_path};
             std::ifstream fasta_input{folder.string() + file_prefix + ebi::vcf::FASTA_EXT};
             std::ifstream fasta_index_input{folder.string() + file_prefix + ebi::vcf::FASTA_EXT + ebi::vcf::INDEX_EXT};
-            CHECK_FALSE(ebi::vcf::assembly_checker::check_vcf_ref(vcf_input, fasta_input, fasta_index_input, outputs));
+            CHECK_FALSE(ebi::vcf::assembly_checker::check_vcf_ref(vcf_input, vcf_path, fasta_input, fasta_index_input, outputs));
         }
+
+        // TODO: make this work in windows
```
cyenyxe: We could wrap this block with a macro so it is still run on Linux and MacOS, as described [here](https://stackoverflow.com/a/8249232).

test/vcf/assembly_checker_integration_test.cpp
```diff
@@ -126,16 +126,19 @@ namespace ebi
...
-            // auto folder = boost::filesystem::path("test/input_files/v4.3/assembly_checker/compressed/failed_compressed_bz2/");
-            // std::string file_prefix = folder.parent_path().filename().string();
-            // std::string vcf_path = folder.string() + file_prefix + ebi::vcf::VCF_BZ2_EXT;
-            // std::ifstream vcf_input{vcf_path};
-            // std::ifstream fasta_input{folder.string() + file_prefix + ebi::vcf::FASTA_EXT};
-            // std::ifstream fasta_index_input{folder.string() + file_prefix + ebi::vcf::FASTA_EXT + ebi::vcf::INDEX_EXT};
-            // CHECK_FALSE(ebi::vcf::assembly_checker::check_vcf_ref(vcf_input, vcf_path, fasta_input, fasta_index_input, outputs));
-        // }
+// TODO: make the next work in windows as well
+#ifdef _WIN32
```
srbcheema1: we could use it like 
```
#ifndef _WIN32
// code goes here
#endif // _WIN32
```

inc/vcf/compression.hpp
```diff
@@ -0,0 +1,81 @@
...
+namespace ebi
+{
+  namespace vcf
+  {
+    /*
+     * method to get compression of a file.
+     * Params:
+     *  source - source_name
+     *  line - vector of first 5 characters or first line(if length of first line < 5) of file.
+     */
```
cyenyxe: Could the comments in this class be structured like Doxygen ones so they are appropriately generated if needed? See [this](https://github.com/EBIvariation/vcf-validator/blob/master/inc/vcf/file_structure.hpp#L260) for an example.

PR #159, opened 2018-09-04
Text report

srbcheema1:
This PR is to add functionality of writing text-report in assembly_checker tool. this report will be similar like the text report in vcf-validator.

### DIFF
--- a/inc/vcf/assembly_report_writer.hpp
+++ b/inc/vcf/assembly_report_writer.hpp
@@ -142,6 +142,11 @@ namespace ebi
 
         ~TextAssemblyReportWriter()
         {
+            file << "Number of matches: " << match_stats.get_num_matches()
+                 << "/" << match_stats.get_num_variants() << std::endl;
+            file << "Percentage of matches: "
+                 << (static_cast<double>(match_stats.get_num_matches()) / match_stats.get_num_variants()) * 100
+                 << "%" << std::endl;
             file.close();
         }
 
--- a/inc/vcf/assembly_report_writer.hpp
+++ b/inc/vcf/assembly_report_writer.hpp
@@ -41,16 +41,22 @@ namespace ebi
     class AssemblyReportWriter
     {
       public:
-        virtual void mismatch(const vcf::RecordCore &record_core) = 0;
-        virtual void match(const vcf::RecordCore &record_core) = 0;
+        virtual ~AssemblyReportWriter(){}
+
+        virtual void match(const vcf::RecordCore &record_core, std::string &match_result) = 0;
+
+        virtual void mismatch(const vcf::RecordCore &record_core, std::string &mismatch_result) = 0;
 
         virtual void finish_report()
         {
-            BOOST_LOG_TRIVIAL(info) << "Number of matches: " << match_stats.get_num_matches() << "/" << match_stats.get_num_variants();
-            BOOST_LOG_TRIVIAL(info) << "Percentage of matches: " << (static_cast<double>(match_stats.get_num_matches()) / match_stats.get_num_variants()) * 100 << "%";
+            BOOST_LOG_TRIVIAL(info) << "Number of matches: "
+                << match_stats.get_num_matches() << "/" << match_stats.get_num_variants();
+            BOOST_LOG_TRIVIAL(info) << "Percentage of matches: "
+                << (static_cast<double>(match_stats.get_num_matches()) / match_stats.get_num_variants()) * 100 << "%";
         }
 
-        bool is_valid_report() {
+        bool is_valid_report()
+        {
             return match_stats.is_assembly_match();
         }
 
@@ -63,14 +69,19 @@ namespace ebi
       public:
         SummaryAssemblyReportWriter(){}
 
-        void mismatch(const vcf::RecordCore &record_core)
+        ~SummaryAssemblyReportWriter()
         {
-            match_stats.add_match_result(false);
+
         }
 
-        void match(const vcf::RecordCore &record_core)
+        void match(const vcf::RecordCore &record_core, std::string & match_result) override
         {
             match_stats.add_match_result(true);
+        }
+
+        void mismatch(const vcf::RecordCore &record_core, std::string &mismatch_result) override
+        {
+            match_stats.add_match_result(false);
         }
     };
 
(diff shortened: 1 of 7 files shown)


src/vcf/assembly_checker.cpp
```diff
@@ -63,9 +73,14 @@ namespace ebi
 
               for (auto &output : outputs ) {
                   if (match_result) {
-                      output->write_match(record_core);
+                      std::string match_result = "Line " + std::to_string(line_num) + ": Reference allele \'"
+                          + reference_sequence + "\' matches with the fasta_sequence";
+                      output->match(record_core,match_result);
                   } else {
-                      output->write_mismatch(record_core);
+                      std::string mismatch_result = "Line " + std::to_string(line_num) + ": Reference allele \'"
```
jmmut: putting the chromosome and position would also be helpful here and in the previous message:

Line 100: (Chromosome 1, position 100) reference allele 'A' does not ...

inc/vcf/assembly_report_writer.hpp
```diff
@@ -63,21 +71,68 @@ namespace ebi
...
+        }
+
+        ~ValidAssemblyReportWriter()
+        {
+            file.close();
+        }
+
+        virtual void match(const vcf::RecordCore &record_core, std::string &match_result) override
         {
             match_stats.add_match_result(true);
```
cyenyxe: This report shouldn't need match statistics, because it only writes the output line.

inc/vcf/assembly_report_writer.hpp
```diff
@@ -41,16 +41,24 @@ namespace ebi
...
+        virtual void write_warning(const vcf::RecordCore &record_core, std::string &warning) = 0;
 
         virtual void finish_report()
         {
-            BOOST_LOG_TRIVIAL(info) << "Number of matches: " << match_stats.get_num_matches() << "/" << match_stats.get_num_variants();
-            BOOST_LOG_TRIVIAL(info) << "Percentage of matches: " << (static_cast<double>(match_stats.get_num_matches()) / match_stats.get_num_variants()) * 100 << "%";
+            BOOST_LOG_TRIVIAL(info) << "Number of matches: "
+                << match_stats.get_num_matches() << "/" << match_stats.get_num_variants();
+            BOOST_LOG_TRIVIAL(info) << "Percentage of matches: "
+                << (static_cast<double>(match_stats.get_num_matches()) / match_stats.get_num_variants()) * 100 << "%";
```
cyenyxe: I think this should apply to the summary report only, not any type of report.

inc/vcf/assembly_report_writer.hpp
```diff
@@ -87,29 +142,33 @@ namespace ebi
...
+            file << "Percentage of matches: "
+                 << (static_cast<double>(match_stats.get_num_matches()) / match_stats.get_num_variants()) * 100
+                 << "%" << std::endl;
             file.close();
         }
 
-        virtual void write_mismatch(const vcf::RecordCore &record_core) override
+        virtual void match(const vcf::RecordCore &record_core, std::string &match_result) override
+        {
+            match_stats.add_match_result(true);
```
cyenyxe: Same comment as in `ValidAssemblyReportWriter`, only the summary should need the statistics.

inc/vcf/assembly_report_writer.hpp
```diff
@@ -87,29 +142,33 @@ namespace ebi
...
+        }
+
+        virtual void mismatch(const vcf::RecordCore &record_core, std::string &mismatch_result) override
         {
             match_stats.add_match_result(false);
-            if(report_type == ebi::vcf::INVALID) {
-                file << record_core.line;
-            }
+            file << mismatch_result << std::endl;
```
cyenyxe: Given that we want errors and warning to look always the same, it would be a better idea to format the message here instead of receiving it as an argument.

inc/vcf/assembly_report_writer.hpp
```diff
@@ -63,21 +71,68 @@ namespace ebi
...
+
+        ~ValidAssemblyReportWriter()
+        {
+            file.close();
+        }
+
+        virtual void match(const vcf::RecordCore &record_core, std::string &match_result) override
         {
             match_stats.add_match_result(true);
+            file << record_core.line;
```
cyenyxe: Even though we aimed to merge both `RecordCore` classes, the idea is to drop this line as an attribute, otherwise the `RecordCore` class won't save any memory compared to the `Record` class. The line can be made an argument for the assembly report writers instead.

src/vcf/assembly_checker.cpp
```diff
@@ -63,9 +73,14 @@ namespace ebi
 
               for (auto &output : outputs ) {
                   if (match_result) {
-                      output->write_match(record_core);
+                      std::string match_result = "Line " + std::to_string(line_num) + ": Reference allele \'"
+                          + reference_sequence + "\' matches with the fasta_sequence";
```
cyenyxe: We will only write mismatches to the report, otherwise we would add a lot of noise and hugely increase its size.

src/vcf/assembly_checker.cpp
cyenyxe: Suggested rewording:

> Line 100: Chromosome 1, position 100, reference allele 'A' does not match the reference sequence, expected 'G'

srbcheema1: done

inc/vcf/assembly_report_writer.hpp
srbcheema1: done that using a local bool variable `is_valid`

inc/vcf/assembly_report_writer.hpp
srbcheema1: done, now the mismatch message is formatted within the function. where as I made seperate functionns for generation of warning messages, as there may be different kind of warnings. right now we have two different warnings

src/vcf/assembly_checker.cpp
srbcheema1: dropped this line.

cyenyxe:
> the reports are always written in the same folder as the input files, or current one if using stdin. a -o parameter would be useful as in the main validator.

+1

> it tries to search angle-bracketed IDs from the CHROM column as-is in the fasta file.

I haven never seen those in a FASTA file, but if that is what the VCF input looks like, then that is what is needs to be checked. The only field that I can recall has a special meaning for angle brackets is ALT (for symbolic alternates).

> in the "valid" report, the header is not written, so the output is not a valid VCF. of course, we are skipping the header entirely so we might need to make some structure changes

Maybe we should disable the "valid" report for now in order to get this PR merged, and enable it in a future one.

PR #161, opened 2018-09-06
Empty warning

srbcheema1:
Vcf validator and assembly checker were showing warning of empty vcf file twice.

```
srb@srb-pc:$ ./vcf_validator < empty.vcf 
[info] Reading from standard input...
[warning] The VCF file provided is empty
[warning] The VCF file provided is empty
[info] Summary report written to : stdin.errors_summary.1536273054595.txt
```

This PR will fix that issue.

### DIFF
--- a/src/vcf/compression.cpp
+++ b/src/vcf/compression.cpp
@@ -100,16 +100,15 @@ namespace ebi
             { { 120, -100 }, ZLIB }
         };
 
-        /*
-         * If the first line of the VCF file is shorter than any magic number string (stored in the variable "types"),
-         * std::equal will cause a segmentation fault, so if the line is shorter than the longest magic number string
-         * (5 characters), then we assume there's no compression.
-         */
-        if (line.size() < 5) {
-            return NO_EXT;
-        }
-
         for (auto & type : types) {
+            /*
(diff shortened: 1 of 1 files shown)


src/vcf/compression.cpp
```diff
@@ -106,9 +106,6 @@ namespace ebi
          * (5 characters), then we assume there's no compression.
          */
         if (line.size() < 5) {
```
cyenyxe: Are we completely sure this is the correct behavior? If the line is 4 characters it could still be compressed some way and have some (invalid) content.

In order to avoid the segmentation fault, we could check it inside the for loop as follows:

```cpp
for (auto & type : types) {
    /*
     * If the first line of the VCF file is shorter than the magic number string (stored in the 
     * variable "types"), the std::equal call below will cause a segmentation fault.
     */
    if (type.first.size() > line.length()) {
        continue;
    }

    if (std::equal(type.first.begin(), type.first.end(), line.begin())) {
        compressed_file_warning(type.second);
        return type.second;
    }

    return NO_EXT;
}
```

srbcheema1: hmm this sounds better.

PR #162, opened 2018-09-06
Valid report

srbcheema1:
This PR is enable the functionality of valid-report. We had disabled valid report in last PR due to the reason that the valid report was not writing `headers` into the output file. So this PR is to fix that issue.

### DIFF
--- a/inc/vcf/assembly_report_writer.hpp
+++ b/inc/vcf/assembly_report_writer.hpp
@@ -112,7 +112,6 @@ namespace ebi
         MatchStats match_stats;
     };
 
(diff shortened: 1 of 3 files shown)


inc/vcf/assembly_report_writer.hpp
```diff
@@ -129,6 +134,11 @@ namespace ebi
             file << vcf_line;
         }
 
+        virtual void write_meta_info(std::string &meta_info_line) override
```
cyenyxe: Now that this code is in place, remove the `TODO` at the beginning of this class.

PR #163, opened 2018-09-06
Option for `outdir` in assembly_checker

srbcheema1:
the assembly checker was missing an option to specify the output directory for the output file. This PR aims to add and `-o` option to assembly_checker

### DIFF
--- a/src/assembly_checker_main.cpp
+++ b/src/assembly_checker_main.cpp
@@ -40,6 +40,7 @@ namespace
             (ebi::vcf::INPUT_OPTION, po::value<std::string>()->default_value(ebi::vcf::STDIN), "Path to the input VCF file, or stdin")
             (ebi::vcf::FASTA_OPTION, po::value<std::string>(), "Path to the input FASTA file; please note that the index file must have the same name as the FASTA file and saved with a .idx extension")
             (ebi::vcf::REPORT_OPTION, po::value<std::string>()->default_value(ebi::vcf::SUMMARY), "Comma separated values for types of reports (summary, text, valid)")
+            (ebi::vcf::OUTDIR_OPTION, po::value<std::string>()->default_value(""), "Output directory")
         ;
 
         return description;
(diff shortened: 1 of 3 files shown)


src/assembly_checker_main.cpp
```diff
@@ -40,6 +40,7 @@ namespace
             (ebi::vcf::INPUT_OPTION, po::value<std::string>()->default_value(ebi::vcf::STDIN), "Path to the input VCF file, or stdin")
             (ebi::vcf::FASTA_OPTION, po::value<std::string>(), "Path to the input FASTA file; please note that the index file must have the same name as the FASTA file and saved with a .idx extension")
             (ebi::vcf::REPORT_OPTION, po::value<std::string>()->default_value(ebi::vcf::SUMMARY), "Comma separated values for types of reports (summary, text)")
+            (ebi::vcf::OUTDIR_OPTION, po::value<std::string>()->default_value(""), "Directory for the output")
```
cyenyxe: Sorry to be a bit picky, but can we replace "directory for the output" with "output directory" in all the tools? It is a much more commonly used expression.

src/assembly_checker_main.cpp
```diff
@@ -137,7 +158,8 @@ int main(int argc, char** argv)
         auto vcf_path = vm[ebi::vcf::INPUT].as<std::string>();
         auto fasta_path = vm[ebi::vcf::FASTA].as<std::string>();
         auto fasta_index_path = fasta_path + ".fai";
-        auto outputs = get_outputs(vm[ebi::vcf::REPORT].as<std::string>(), vcf_path);
+        auto outdir = get_output_path(vm[ebi::vcf::OUTDIR].as<std::string>(), vcf_path);
+        auto outputs = get_outputs(vm[ebi::vcf::REPORT].as<std::string>(), outdir);
```
jmmut: it would be nice to log which is the output path, just as the validator does:
[info] Summary report written to : ../empty.vcf.errors_summary.1536308696324.txt

we would need to do it for the text report and the valid report, so the valid PR needs to be after this PR

PR #166, opened 2018-09-13
include db_name in error message

srbcheema1:
this PR is to include path of DB file during the file not opened error. as per the issue #142

### DIFF
--- a/src/validator_main.cpp
+++ b/src/validator_main.cpp
@@ -134,9 +134,16 @@ namespace
                 std::string errortype = (out == ebi::vcf::SUMMARY) ? "errors_summary" : "errors";
                 std::string filename = input + "." + errortype + "." + std::to_string(timestamp) + "." + filetype;
                 boost::filesystem::path file{filename};
-                if (boost::filesystem::exists(file)) {
-                    throw std::runtime_error{"Report file already exists on " + filename + ", please delete it or rename it"};
+
+                try {
+                    if (boost::filesystem::exists(file)) {
+                        throw std::runtime_error{"Report file already exists on " + filename + ", please delete it or rename it"};
+                    }
+                } catch (boost::filesystem::filesystem_error const & er) {
+                    // commented it to let it be caught in RepportWriter
+                    // throw std::runtime_error{"Permissions denied, can't access the location " + filename};
                 }
+
                 if (out == ebi::vcf::DATABASE) {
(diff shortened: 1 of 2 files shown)


src/validator_main.cpp
```diff
@@ -134,9 +134,15 @@ namespace
...
                 boost::filesystem::path file{filename};
-                if (boost::filesystem::exists(file)) {
-                    throw std::runtime_error{"Report file already exists on " + filename + ", please delete it or rename it"};
+
+                try {
+                    if (boost::filesystem::exists(file)) {
+                        throw std::runtime_error{"Report file already exists on " + filename + ", please delete it or rename it"};
+                    }
+                } catch (boost::filesystem::filesystem_error const & er) {
+                    throw std::runtime_error{"Permissions denied, can't access the location " + filename};
```
jmmut: I'm not sure about this catch and this message, the error could be another one. either test the permissions and keep the message, or remove the catch and keep only the one in odb_report.

src/vcf/odb_report.cpp
```diff
@@ -53,7 +53,7 @@ namespace ebi
                 }
             }
         } catch (const odb::exception& e) {
-            throw std::runtime_error{std::string{"ODB report: Can't initialize database: "} + e.what()};
+            throw std::runtime_error{std::string{"ODB report: Can't initialize database: " + db_name + " "} + e.what()};
```
jmmut: > std::runtime_error{"ODB report: Can't initialize database file \"" + db_name + "\": " + e.what()};

src/validator_main.cpp
```diff
@@ -134,9 +134,16 @@ namespace
...
-                if (boost::filesystem::exists(file)) {
-                    throw std::runtime_error{"Report file already exists on " + filename + ", please delete it or rename it"};
+
+                try {
+                    if (boost::filesystem::exists(file)) {
+                        throw std::runtime_error{"Report file already exists on " + filename + ", please delete it or rename it"};
+                    }
+                } catch (boost::filesystem::filesystem_error const & er) {
+                    // commented it to let it be caught in RepportWriter
+                    // throw std::runtime_error{"Permissions denied, can't access the location " + filename};
```
jmmut: after comparing how it behaves in develop and here, the develop behaviour is better, as the boost exception explains better the problem than we. please revert every change in this file.

PR #167, opened 2018-09-16
warn for unused parameters

### LINKED ISSUE #165: Warn if there were unused parameters
In the (I hope) unlikely case that someone is using the parameters incorrectly and some parameters were not used, it would be nice to show a warning with the unused parameters:

```
$ ./vcf_validator -i a.vcf -r summary, text
[warn] unused parameters: "text"
```

srbcheema1:
This PR is to fix #165. vcf_validator will show a warning if any of the parameters went unused or unrecognised. 
```
$ ./vcf_validator -i a.vcf -r summary, text
[warning] unused parameter: text
[info] Summary report written to : stdin.errors_summary.1537119684362.txt

```

### DIFF
--- a/inc/util/cli_utils.hpp
+++ b/inc/util/cli_utils.hpp
@@ -24,7 +24,6 @@
 #include <boost/filesystem/operations.hpp>
 
 #include "util/logger.hpp"
-#include "vcf/validator.hpp"
 
 namespace ebi
 {
--- a/inc/util/cli_utils.hpp
+++ b/inc/util/cli_utils.hpp
@@ -21,6 +21,7 @@
 #include <vector>
 
 #include <boost/program_options.hpp>
+#include <boost/filesystem/operations.hpp>
 
 #include "util/logger.hpp"
 #include "vcf/validator.hpp"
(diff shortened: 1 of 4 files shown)


src/validator_main.cpp
```diff
@@ -160,15 +160,20 @@ int main(int argc, char** argv)
 
     po::options_description desc = build_command_line_options();
     po::variables_map vm;
-    po::store(po::parse_command_line(argc, argv, desc), vm);
-    po::notify(vm);
+    po::parsed_options parsed = po::command_line_parser(argc, argv).options(desc).allow_unregistered().run();
+    po::store(parsed,vm);
+    std::vector<std::string> unused_args = collect_unrecognized(parsed.options, po::include_positional);
+
+    for (auto a: unused_args) {
```
jmmut: use descriptive names

src/validator_main.cpp
```diff
@@ -160,15 +160,20 @@ int main(int argc, char** argv)
 
     po::options_description desc = build_command_line_options();
     po::variables_map vm;
-    po::store(po::parse_command_line(argc, argv, desc), vm);
-    po::notify(vm);
+    po::parsed_options parsed = po::command_line_parser(argc, argv).options(desc).allow_unregistered().run();
```
jmmut: what about extracting this behaviour to a function `build_variables_map`? including the check for unused args

src/validator_main.cpp
srbcheema1: what about unused_parameters or unredognised_options ?

jmmut: oh, sorry, I was referring to the `a` variable, although `unrecognised_parameters` is also better that `unused_args`

inc/util/cli_utils.hpp
```diff
@@ -0,0 +1,73 @@
...
+#define UTIL_CLI_UTILS_HPP
+
+#include <string>
+#include <vector>
+
+#include <boost/program_options.hpp>
+#include <boost/filesystem/operations.hpp>
+
+#include "util/logger.hpp"
+#include "vcf/validator.hpp"
```
jmmut: is this include needed? Ideally the `util` namespace won't depend on the `vcf` namespace

PR #168, opened 2018-09-21
EVA-1256 Report duplicate samples

jmmut:
Example output:

```
$ ./bin/vcf_validator -i ../../test/input_files/v4.3/failed/failed_body_sample_011.vcf -o /tmp
[info] Summary report written to : /tmp/failed_body_sample_011.vcf.errors_summary.1537521390924.txt

$ cat /tmp/failed_body_sample_011.vcf.errors_summary.1537521390924.txt
According to the VCF specification, the input file is not valid
Error: The header line has 3 repeated sample names: [HG00096, HG00097, HG00096]. This occurs 1 time(s), first time in line 3.
Warning: A valid 'reference' entry is not listed in the meta section. This occurs 1 time(s), first time in line 4.
Warning: Chromosome/contig '1' is not described in a 'contig' meta description. This occurs 1 time(s), first time in line 4.
```


src/vcf/store_parse_policy.cpp
```diff
@@ -112,6 +115,28 @@ namespace ebi
...
     {
         state.set_samples(m_grouped_tokens);
+        check_each_sample_appears_only_once(state);
+    }
+    
+    void StoreParsePolicy::check_each_sample_appears_only_once(ParsingState const &state) const
+    {
+        using util::operator<<;
+        std::set<std::string> unique_samples;
+        std::vector<std::string> repeated_sample_names;
```
cyenyxe: When a sample is a triplicate it is reported twice, which is somewhat confusing. I think this would better be a set instead of a vector.

jmmut: I did that on purpose, suppose you have a big list (thousands of samples), and it says 1 duplicate. you remove it, and after validating, it reports the same duplicate again. I think that would be way more confusing than having the reported list with repeated names.

jmmut: message changed to 
> "The header line has 3 repeated sample names. Please remove these names from the header line and the data lines: HG00096 (2 times), HG00097."

PR #169, opened 2018-09-21
improve cmake with dynamic build to have more debug symbols

jmmut:


CMakeLists.txt
```diff
@@ -54,19 +54,29 @@ else ()
 endif ()
 
 
+# Even in "Debug" build type, static linking prevents storing some symbol information and makes debugging more difficult.
+# use this flag to avoid `$1 = <incomplete type>` messages
+if (NOT KEEP_SYMBOL_INFO)
+  set (KEEP_SYMBOL_INFO false CACHE BOOL "Even in Debug build type, static linking prevents storing some symbol information and makes debugging more difficult. Use this flag to avoid `$1 = <incomplete type>` messages" FORCE)
```
srbcheema1: I think `FORCE` is useless here as the set command in `if` condition will never be executed if it is already in `CMakeCache` .

CMakeLists.txt
```diff
@@ -54,19 +54,29 @@ else ()
...
-  set (CMAKE_EXE_LINKER_FLAGS "-static")
-  set (CMAKE_FIND_LIBRARY_SUFFIXES .a)
-  set (CMAKE_EXE_LINK_DYNAMIC_C_FLAGS)       # remove -Wl,-Bdynamic
-  set (CMAKE_EXE_LINK_DYNAMIC_CXX_FLAGS)
-  set (CMAKE_SHARED_LIBRARY_C_FLAGS)         # remove -fPIC
-  set (CMAKE_SHARED_LIBRARY_CXX_FLAGS)
-  set (CMAKE_SHARED_LIBRARY_LINK_C_FLAGS)    # remove -rdynamic
-  set (CMAKE_SHARED_LIBRARY_LINK_CXX_FLAGS)
   set (Boost_USE_STATIC_LIBS ON)             # only find static libs
+  if (NOT KEEP_SYMBOL_INFO)
```
srbcheema1: would it be cool to say `if(NOT CMAKE_BUILD_TYPE STREQUAL "Debug")`

srbcheema1:
I think `FORCE` keyword can also be dropped from https://github.com/EBIvariation/vcf-validator/blob/6e68167bcba5a15d64208aa8ddfee7b9c1d3508d/CMakeLists.txt#L82 
as stated in cmake documentataion 
> Normally, set(…CACHE…) creates cache variables, but does not modify them. If FORCE is specified, the value of the cache variable is set, even if the variable is already in the cache. This should normally be avoided, as it will remove any changes to the cache variable’s value by the user.

so only `set(...CACHE...)` would be enough as per our requirement.

CMakeLists.txt
```diff
@@ -54,19 +54,29 @@ else ()
...
+      # Static Build Extra Flags for Linux Static Build
```
srbcheema1: intendataion prob

CMakeLists.txt
jmmut: the thing is I think the problem is linking the system libs statically. if we use the condition you suggest, it won't be possible to have a Debug binary with static linking.

srbcheema1: do we need `debug binary using static linking` ? as you suggested that static linking is creating some problem, so why shouldn't we drop that case `debug + static linking`.

srbcheema1:
https://github.com/EBIvariation/vcf-validator/blob/a0ed5fbb98a9f96476c27f3cee244008c44eac0a/CMakeLists.txt#L90-L93

these lines could be converted into a single line.

> set (CMAKE_BUILD_TYPE  "RelWithDebInfo" CACHE STRING "Choose the type of build, options are: None(CMAKE_CXX_FLAGS or CMAKE_C_FLAGS used) Debug Release RelWithDebInfo MinSizeRel.")

It will serve the same purpose.

CMakeLists.txt
jmmut: I did it this way to keep it independent, and also being able to compare debug+dynamic and debug+static. Not really sure what's the best here

tcezard:
Not needed: closing

Issue #172, opened 2018-11-02
Ensure compressed FASTA files are handled properly


jmmut:
I expect the assembly checker won't be able to use compressed FASTAs, even having the index file (.fai), so we have to manually check that it fails on compressed FASTAs and then put a check in the code to throw an exception if the FASTA is compressed, and stop the program with a clear error message.

Otherwise we will read rubbish and it will say 0% matches the reference, or not finding the contigs.

tcezard:
This has been done as part of EVA-1731

PR #173, opened 2018-11-16
checks for compressed files

srbcheema1:
This PR is to solve issue #172 
This will add functionality to verify a fasta, fasta_index and assembly_report file to be uncompressed.


tcezard:
This has been done as part of EVA-1731

PR #174, opened 2018-11-16
Travis

srbcheema1:
This PR aims to simplify the travis build. earlier the travis used to run release-builds for non-tagged branches too. After this PR the travis will run release builds only for tagged commits saving lot of time on every build.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,8 +12,8 @@ cache:
 jobs:
   include:
 
+    - stage: "Compile and Test"
     # OS X Clang
-    - stage: "Supported Platforms and Compilers"
       os: osx
       osx_image: xcode9.3
       env: C_COMPILER=clang CXX_COMPILER=clang++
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -12,8 +12,9 @@ cache:
 jobs:
   include:
 
+    - stage: "Supported Platforms and Compilers"
```
cyenyxe: Maybe this should be "Compile and Test", otherwise it is going to look really weird when is it displayed along "Release builds".

.travis.yml
```diff
@@ -126,21 +127,31 @@ jobs:
...
+            - bin/vcf_debugulator_macos
+            - bin/vcf_assembly_checker_macos
         skip_cleanup: true
         on:
             repo: EBIvariation/vcf-validator
             tags: true
-      os: linux
+
+      # Linux Release
+    - os: linux
```
jmmut: isn't this just another stage? it's good to add "tag IS present" but the other changes seem confusing to me, the stage named "release builds" only builds mac, and this linux stage is still a separate stage. It looks like the GUI just adds the message but it's not grouped in the code

jmmut: did you notice this? after we discuss it, I'm ok merging the code

cyenyxe: I was also very confused by it, but it seems to be just stupid syntax, look at [this example](https://docs.travis-ci.com/user/build-stages/).

srbcheema1: Yepp the syntax is stupid. but I just followed the pattern that they have written without going into why did they choose this syntax. I too was unhappy with this pattern
```
jobs:
  include:
    - stage: test
      script: ./test 1
    - # stage name not required, will continue to use `test`
      script: ./test 2
    - stage: deploy
      script: ./deploy
```
there test1 and test2 are sharing a common stage. the stage's scope continues till next stage is declared.

srbcheema1: `this linux stage is still a separate stage` no this is not a seperate. from code it seems to be. but unfortunately it is like it is shown in GUI. even I am unhappy with this syntax. But they kept it like this so we have no other option.

`It looks like the GUI just adds the message but it's not grouped in the code`. Gui is behaving as it is supposed to behave, just the code is behaving differently from what it `seems`.

PR #175, opened 2018-12-04
EVA-1406 Support custom IDs in the ALT column

jmmut:
Previously, custom ALT ids were only allowed as subtypes (e.g.
<DEL:MY_ID>). After a discussion with the spec, custom ids are
also allowed as main types (e.g. <MY_ID>).

This means that the predefined types should only be at the beginning
of an id if it has subtypes, i.e. contains ':'.


src/vcf/meta_entry.cpp
```diff
@@ -129,17 +129,18 @@ namespace ebi
...
+        // Check ID main type (prefix before ':') is "DEL" | "INS" | "DUP" | "INV" | "CNV"
+        auto main_type_position_end = id_field.find(':');
+        bool colon_present = main_type_position_end != std::string::npos;
+        if (colon_present) {
+            auto main_type = id_field.substr(0, main_type_position_end);
+            if (main_type != DEL
+                && main_type != INS
+                && main_type != DUP
+                && main_type != INV
+                && main_type != CNV) {
```
cyenyxe: Now that you're into this, could you make it support BND too? (see section 1.4.5 of the specification)

src/vcf/vcf.ragel
```diff
@@ -514,10 +514,11 @@
     record_alt_snv  = bases ;
     
     ## Indel alternates can be represented by standardized prefixes or an asterisk
-    record_alt_indel= '<DEL>' | '<INS>' | '<DUP>' | '<INV>' | '<CNV>' | '<DUP:TANDEM>' | '<DEL:ME:' (alnum)+ '>' | '<INS:ME:' (alnum)+ '>' | '*';
+    record_alt_prefix = '<DEL>' | '<INS>' | '<DUP>' | '<INV>' | '<CNV>' | '<DUP:TANDEM>' | '<DEL:ME:' (alnum)+ '>' | '<INS:ME:' (alnum)+ '>';
+    record_alt_indel = record_alt_prefix | '*';
 
     ## Other alternates can be any identifier surrounded by < > symbols
-    record_alt_other= ('<' alt_id '>') - ('<DEL>' | '<INS>' | '<DUP>' | '<INV>' | '<CNV>' | '<DUP:TANDEM>' | '<DEL:ME:' (alnum)+ '>' | '<INS:ME:' (alnum)+ '>');
+    record_alt_other = ('<' alt_id '>') - record_alt_prefix;
```
cyenyxe: `alt_id` should not accept characters that are not valid contig names. Could you please check that is the case? I see so many symbols in one of the passing tests that I'm almost scared :fearful:

jmmut: the spec says for ALT IDs:
> ID values are case sensitive strings and must not contain whitespace or angle brackets

and 
> (String; no whitespace, commas, or angle-brackets are permitted in the ID String itself)

and for contigs:
> Valid contig names must follow the reference sequence names allowed by the SAM format (`"[!-)+-<>-~][!-~]*"`) excluding the characters `"<>[]:*"` to avoid clashes with symbolic alleles and breakend notation.  The contig names must not use a reserved symbolic allele name.

so, are you talking about modifying the spec?

(I don't really understand those strings of the SAM format. Are those regex? character lists?)

jmmut: ok, interpreting the contig definition `"[!-)+-<>-~][!-~]*"` as a regex (`!-)` means the range from `!` to `)`) and excluding `"<>[]:*"`, then the set of valid ALT IDs and the set of valid contigs, don't match according to the spec (space is forbidden in both):
- chars that can't appear in ALT: `,<>`
- chars that can't appear in the first letter of a contig: `*=<>[]:`
- chars that can't appear in subsequent letters of a contig: `<>[]:*`

(space is forbidden in both, the rest of the characters are printable)

from this we can see that:
- the next chars can appear in a contig but not in an ALT: `<>,`
- the next chars can appear in an ALT but not in the first letter of a contig: `*=[]:`
- the next chars can appear in an ALT but not in subsequent letters of a contig: `*[]:`

jmmut: TL;DR according to the spec, "alt_id should not accept characters that are not valid contig names" is not true. Shall we keep the validator as it is, or shall we raise this to the spec?

PR #176, opened 2018-12-05
T2D-268 Add support for retrieving reference sequence from ENA

Zhicheng-Liu:
With this change, the assembly checker can access compressed fasta, fasta
files in a remote location and download contig reference sequence on
demand from ENA API.


inc/util/file_utils.hpp
```diff
@@ -37,6 +43,25 @@ namespace ebi
         }
     }
 
+    inline void open_local(std::ifstream & input,
```
jmmut: I don't see the point of this function, why not use open_file directly? if you want to highlight it's a local file, you can rename open_file to open_local_file. If your concern is const-correctness, you can add const to the old function

inc/vcf/fasta.hpp
```diff
@@ -0,0 +1,74 @@
...
+
+#include <boost/iostreams/filtering_stream.hpp>
+#include "bioio/bioio.hpp"
+#include "util/curl_easy.hpp"
+
+namespace ebi
+{
+  namespace vcf
+  {
+    namespace fasta
```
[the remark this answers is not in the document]
srbcheema1: I would say it would be better to take it out of `vcf` namespace. similar like `assembly_report`.

src/vcf/fasta.cpp
```diff
@@ -0,0 +1,131 @@
...
+}
+
+std::shared_ptr<std::istream> create_input_stream(const std::string &path) {
+  if ( !ebi::util::is_remote_url(path) ) {
+    std::shared_ptr<std::ifstream> fstream(new std::ifstream());
+    ebi::util::open_local(*fstream, path, std::ifstream::binary);
+    return create_decompressed_stream(std::static_pointer_cast<std::istream>(fstream));
+  }
+
+  std::shared_ptr<std::stringstream> sstream(new std::stringstream());
```
srbcheema1: any particular reason for using sstream instead of ifstream ?

src/vcf/fasta.cpp
```diff
@@ -0,0 +1,131 @@
...
+  return std::static_pointer_cast<std::istream>(ss);
+}
+
+std::shared_ptr<std::istream> create_decompressed_stream(std::shared_ptr<std::istream> stream) {
+  std::vector<char> line;
+  ebi::vcf::get_magic_num(*stream, line);
+  std::string file_extension = ebi::vcf::get_compression("", line);
+
+  if ( file_extension == ebi::vcf::GZ ) {
+    return create_gunzipped_stream(stream);
```
srbcheema1: can we use `ebi::vcf::create_uncompressed_stream` directly here ?

src/vcf/fasta.cpp
```diff
@@ -0,0 +1,131 @@
...
+
+  return std::static_pointer_cast<std::istream>(ss);
+}
+
+std::shared_ptr<std::istream> create_decompressed_stream(std::shared_ptr<std::istream> stream) {
+  std::vector<char> line;
+  ebi::vcf::get_magic_num(*stream, line);
+  std::string file_extension = ebi::vcf::get_compression("", line);
+
+  if ( file_extension == ebi::vcf::GZ ) {
```
srbcheema1: I am afraid if uncompressing a fasta file would allow random reads that we do in `bioio` library

src/vcf/assembly_checker.cpp
```diff
@@ -22,10 +22,14 @@ namespace ebi
   {
     namespace assembly_checker
     {
+      bool process_vcf_records(std::istream & vcf_input,
+                               std::shared_ptr<ebi::vcf::fasta::IFasta> & fasta,
+                               const std::string & assembly_report,
+                               std::vector<std::unique_ptr<ebi::vcf::AssemblyCheckReportWriter>> & outputs);
+
```
srbcheema1: I think we can either shift this declaration to hpp file or we can move `process_vcf_ref` declaration here. I would go with moving `process_vcf_ref` to this file. we may keep only those declarations in hpp which are used elsewhere in other files. what do you say @cyenyxe @jmmut

inc/util/curl_easy.hpp
```diff
@@ -0,0 +1,49 @@
...
+
+#include <curl/curl.h>
+
+namespace ebi
+{
+  namespace util
+  {
+    namespace curl
+    {
+      class Easy final
```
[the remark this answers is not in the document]
Zhicheng-Liu: I created this class from scratch. But the implementation was following some of the examples at libcurl(https://curl.haxx.se/libcurl/c/example.html).

inc/vcf/fasta.hpp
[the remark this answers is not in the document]
Zhicheng-Liu: I agree `vcf` is not a good place to put it.  There are a few other headers in `vcf` folder that are used not only for VCFs, they are more like utilities, e.g.: `compression.hpp` and `string_constants.hpp`. I would prefer to put the `fasta` class somewhere together with other reference/assembly/contig related classes.

src/vcf/assembly_checker.cpp
Zhicheng-Liu: I would like leaving only one function in the `assembly_checker` header with current implementation and forward declare other functions. I didn't do it because I did not want to mix this kind of refactoring with a feature development.

src/vcf/fasta.cpp
Zhicheng-Liu: If you are referring to the current implementation of `bioio` library I did not see it could handle compressed fasta file/stream. If have missed something, please point it out @srbcheema1. The point of decompressing `gz` stream first is to allow random access later, which by its nature a `gz` stream does not allow.

src/vcf/fasta.cpp
Zhicheng-Liu: In `create_gunzipped_stream` function it is calling `ebi::vcf::create_uncompressed_stream`. But for a `gz` stream, it needs to go futher on that and decompress it to allow random access.

src/vcf/fasta.cpp
Zhicheng-Liu: First it has to be a stream for both input and output because it is getting written to when downloading and re-used laster as an input to assembly checker. Definitly not `ifstream`, `fstream` could do. But I don't want to suggest this stream is associated with a file, it is merely a `string`.

inc/vcf/fasta.hpp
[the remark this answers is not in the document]
jmmut: for now I would say to put your classes related to fastas in inc/fasta and src/fasta, and in a later PR we can refactor the classes that are already in the repo.

This way we do little changes but still we lay down the correct structure, and later we just have to align the rest

src/vcf/fasta.cpp
jmmut: ok, maybe I'm not aware of all the conversation but now I realised this PR adds 3 features (compressed fastas, remote fastas, ENA API), while the jira ticket asks only for the ENA API. Did I miss anything?

jmmut: because I can see that the design was changed to load even decompressed fastas in memory. The point about having fasta indexes is to not load the whole fasta in memory. have you tried validating a vcf using a big decompressed fasta?

src/vcf/fasta.cpp
```diff
@@ -0,0 +1,131 @@
...
+
+    // This contig is not downloaded, try download it from ENA.
+    if (contig.size() < 6) { // don't trust short contig names.
+      return "";
+    }
+
+    // Download this contig
+    std::string url = ebi::vcf::ENA_API_URL + contig + "&" + ebi::vcf::ENA_API_FASTA_FORMAT;
+    std::stringstream ss;
+    ebi::util::open_remote(ss, url);
```
jmmut: so this downloads the whole contig from ENA into memory? I could be ok with this for now, to have the basic workings, but I think we can optimise this to download chunks, and free the chunks after they no longer have the requested region, as the VCFs should be sorted

src/util/curl_easy.cpp
```diff
@@ -0,0 +1,77 @@
...
+#include "util/curl_easy.hpp"
+
+size_t
+StringWriterCallback(void *contents, size_t size, size_t nmemb, void *buffer) {
+  ((std::string*)buffer)->append((const char*)contents, size * nmemb);
+  return size * nmemb;
+}
+
+size_t
+StreamWriterCallback(void *contents, size_t size, size_t nmemb, void *stream) {
```
jmmut: we use camelCase for functions. At first sight I thought this was a constructor

src/vcf/fasta.cpp
srbcheema1: ohh my bad. I havent seen implementation of open_remote. I thought it could be like first downloading the file and then opening.
But I think downloading and then opening seems good idea in case the file is too large. why to load whole of it into the memory.

src/vcf/fasta.cpp
srbcheema1: what about downloading it into a file on hard-drive instead of downloading into a memory. this would also help in caching that file for later use. in case we need the same chunk in the next run.

src/vcf/assembly_checker.cpp
```diff
@@ -36,21 +40,94 @@ namespace ebi
...
+          std::string reference_accession;
+          std::vector<std::string> contigs;
+
+          for (size_t line_num = 1; util::readline(vcf_input, vector_line).size() != 0; ++line_num) {
+            std::string line{vector_line.begin(), vector_line.end()};
+            if (!boost::starts_with(line, "##")) {
+              break;
+            }
+            for (auto &output : outputs) {
+              output->write_meta_info(line);
```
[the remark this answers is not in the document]
Zhicheng-Liu: This is to ensure that the current metadata line is written to outputs. Otherwise the stream has be "rewinded" back to the beginning. On some platforms random seek may not be available for the input. I had to revert it back in change 880f5a9 .

inc/vcf/fasta.hpp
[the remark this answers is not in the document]
Zhicheng-Liu: For now, I have moved the `fasta.h` and `fasta.cpp` to `fasta` subfolders.

src/vcf/fasta.cpp
Zhicheng-Liu: The new implementation is now based on storing the downloaded sequence in a tmp file and delete the tmp file once it is done with.

src/vcf/fasta.cpp
Zhicheng-Liu: The new implementation has removed the functionalities of reading compressed or remote fasta files. It is now only getting reference sequences from ENA API.

src/vcf/fasta.cpp
Zhicheng-Liu: As mentioned above, the sequence will be written to a tmp file which later will be removed. If we want to cache those reference sequences for later use, there should be limit or a curated list to work with.

src/util/curl_easy.cpp
Zhicheng-Liu: Fixed.

inc/fasta/fasta.hpp
```diff
@@ -0,0 +1,76 @@
...
+      public:
+        virtual std::string sequence(const std::string& contig, const size_t pos, const size_t length) = 0;
+        virtual size_t count(const std::string& contig) const = 0;
+      };
+
+      class FileBasedFasta : public IFasta
+      {
+      public:
+        FileBasedFasta(const std::string& fasta_path, const std::string& fasta_index_path);
+        virtual ~FileBasedFasta(){}
```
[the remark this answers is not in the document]
Zhicheng-Liu: True. What a miss!

test/vcf/assembly_checker_integration_test.cpp
```diff
@@ -39,9 +39,10 @@ namespace ebi
             std::string file_prefix = folder.parent_path().filename().string();
             std::string vcf_path = folder.string() + file_prefix + ebi::vcf::VCF_EXT;
             std::ifstream vcf_input{vcf_path};
-            std::ifstream fasta_input{folder.string() + file_prefix + ebi::vcf::FASTA_EXT};
-            std::ifstream fasta_index_input{folder.string() + file_prefix + ebi::vcf::FASTA_EXT + ebi::vcf::INDEX_EXT};
-            CHECK(ebi::vcf::assembly_checker::check_vcf_ref(vcf_input, vcf_path, fasta_input, fasta_index_input, assembly_report, outputs));
+            std::string fasta_input{folder.string() + file_prefix + ebi::vcf::FASTA_EXT};
+            std::string fasta_index_input{folder.string() + file_prefix + ebi::vcf::FASTA_EXT + ebi::vcf::INDEX_EXT};
```
cyenyxe: Could these be renamed as `fasta_path` and `fast_index_path`, to match the naming criteria for the VCF file? Otherwise it gets a bit confusing whether they are paths or input.

CMakeLists.txt
```diff
@@ -43,6 +43,7 @@ elseif (OSX AND EXISTS "${PROJECT_SOURCE_DIR}/osx_dependencies")
 elseif (WINDOWS AND EXISTS "${PROJECT_SOURCE_DIR}/windows_dependencies")
   include_directories (lib/sqlite)            # Windows specific
   include_directories (windows_dependencies)
+  include_directories (windows_dependencies/curl/include)
   set (EXT_LIB_PATH "${PROJECT_SOURCE_DIR}/lib/windows_specific" CACHE STRING "Absolute path to libraries")
```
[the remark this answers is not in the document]
Zhicheng-Liu: For Windows, yes, the `EXT_LIB_PATH` has been set to `lib/windows_specific` and linking against `odb.lib` and `odb-sqlite.lib` from that directory. While in `install_dependencies.bat` we are downloading these two libraries but they seem not being used.

inc/util/string_utils.hpp
```diff
@@ -93,7 +97,17 @@ namespace ebi
         }
         return eol;
     }
+
+    inline bool is_remote_url(const std::string& input)
+    {
+      return boost::starts_with(input, ebi::vcf::HTTP) ||
+             boost::starts_with(input, ebi::vcf::HTTPS) ||
+             boost::starts_with(input, ebi::vcf::FTP) ||
+             boost::starts_with(input, ebi::vcf::FTPS);
```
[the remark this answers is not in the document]
Zhicheng-Liu: Yes, we do. The callers of `ebi::util::is_remote_url` are checking whether the input string is of URL form. So we have to check the common protocols at the least.

src/vcf/assembly_checker.cpp
```diff
@@ -36,21 +43,64 @@ namespace ebi
...
+              }
+            }
+
+            if (!boost::starts_with(line, "##")) {
+              break;
+            }
+
+            if (boost::starts_with(line, "##reference")) {
+              reference_accession = get_reference_accession(line);
+              continue;
```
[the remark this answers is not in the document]
Zhicheng-Liu: For the parsing of `##reference` and `##contig` lines, we can't consolidate them because they follow different format. Once we have accessions from either or both of them, they end up in the same way in the same place. So no much further consolidation we can do here really imho.

inc/util/string_utils.hpp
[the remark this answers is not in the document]
cyenyxe: But given our use case, do we need to implement such a generic method, or could it be `is_http_url`? (maybe not the best name because it also checks against HTTPS)

src/vcf/assembly_checker.cpp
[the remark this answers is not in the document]
cyenyxe: > Once we have accessions from either or both of them, they end up in the same way in the same place

This a good point and should address part of my concerns.

I still think there should be a brief comment indicating that if 'reference' is present, we won't use the 'contig'.

Zhicheng-Liu:
@jmmut I have fixed the issues you mentioned above:
>     * when only the vcf is passed to the program (`vcf_assembly_checker -i my.vcf`):
>       
>       * empty files are created if the contig is not found in ENA
>       * segmentation fault occurs, leaving the files in the folder. (segfault freeing the IFasta in the main function).
>       * when no `##reference` nor `##contig` metadata rows are present, No RemoteContig download is triggered, and a segfault occurs
This was introduced with the new implementation. I did not test it properly. This should have now been fixed by 8ef6838. A test case is also added to cover this scenario. 

>     * when a remote contig is downloaded, before writing to a file it is loaded in memory in a stringstream
Good catch. My bad again. The new implementation was supposed to be fixing such issue. There were starts and stops, so I must have forgotten something at some point. Anyway it is now fixed in e5ceed1.

inc/util/string_utils.hpp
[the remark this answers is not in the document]
Zhicheng-Liu: This method is used in two different ways:
1. to ignore the remote files when looking for the reference accession through `##reference` tag value. In some VCFs, the `##reference` tag value is a remote file location.
2. to check that the URL has one of these four protocols when trying to download something through libcurl utility.
Although we only use `https` for ENA API, I still think we should keep all of these given the usage 1 mentioned above.

src/vcf/assembly_checker.cpp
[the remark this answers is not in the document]
Zhicheng-Liu: The current implementation is using both `reference` and `contig` accessions. Usually the `reference` accession is not supported on ENA side, but the `contig` accessions are, e.g. human reference genome.

src/fasta/fasta.cpp
```diff
@@ -30,108 +30,118 @@
...
+{
+    if (contigs.count(contig) == 0) {
+        contigs[contig].reset(new ebi::vcf::fasta::ContigFromENA(contig));
 
-    // This contig is not downloaded, try download it from ENA.
-    if (contig.size() < 6) { // don't trust short contig names.
-      return "";
-    }
+        // This contig is not downloaded, try download it from ENA.
+        if (contig.size() < 6) { // don't trust short contig names.
```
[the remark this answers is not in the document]
Zhicheng-Liu: The intention was not to download the referene sequence for a chromosome like `chr22`, but this will prevent downloading for contigs with short accessions.
After a further look, this is not absolutely necessary as we will get an empty reference sequence if the short name is not supported. I have hence removed this check in dc4dda3 .

inc/vcf/string_constants.hpp
```diff
@@ -198,6 +199,15 @@ namespace ebi
     const std::string ZIP = ".zip";
     const std::string ZLIB = ".z";
 
+    // Network protocols
+    const std::string HTTP = "http://";
+    const std::string HTTPS = "https://";
+    const std::string FTP = "ftp://";
+    const std::string FTPS = "ftps://";
+
+    // ENA REST API for remote fasta access
```
cyenyxe: We recently found out about a new ENA API that retrieves the requested version of a contig, instead of just the latest one. Everyone in the EVA team should know which is it, could you please ask and replace it? Otherwise I think this is ready to merge.

PR #177, opened 2018-12-05
EVA-1406 Support custom IDs in the ALT column (and forbid custom SVTYPES)

jmmut:
Also, minor fixes related to ALT ID and the BND type

### DIFF
--- a/inc/vcf/error.hpp
+++ b/inc/vcf/error.hpp
@@ -163,7 +163,7 @@ namespace ebi
                 value{value},
                 expected_value{expected_value} {
                     if (error_fix == ErrorFix::RECOVERABLE_VALUE && (value.empty() || expected_value.empty())) {
-                        throw std::invalid_argument("An error with recoverable meta defintion must provide non-empty field and expected values");
+                        throw std::invalid_argument("An internal error occurred: Errors with recoverable meta definition must provide non-empty field and expected values");
                     }
                 }
         virtual ~MetaSectionError() override { }
@@ -306,7 +306,7 @@ namespace ebi
                 field{field},
                 expected_value{expected_value} {
                     if (error_fix == ErrorFix::RECOVERABLE_VALUE && expected_value.empty()) {
-                        throw std::invalid_argument{"The program had an internal error: An error with recoverable value must provide a non-empty expected value"};
+                        throw std::invalid_argument{"An internal error occurred: Errors with recoverable value must provide a non-empty expected value"};
                     }
                 }
         virtual ~InfoBodyError() override { }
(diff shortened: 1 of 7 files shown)


inc/vcf/error.hpp
```diff
@@ -150,7 +150,18 @@ namespace ebi
...
+                         const std::string &detailed_message,
+                         ErrorFix error_fix = ErrorFix::IRRECOVERABLE_VALUE,
+                         const std::string &value = "",
+                         const std::string &expected_value = "")
+                : Error{line, message, detailed_message},
+                error_fix{error_fix},
+                value{value},
+                expected_value{expected_value} {
                     if (error_fix == ErrorFix::RECOVERABLE_VALUE && (value.empty() || expected_value.empty())) {
                         throw std::invalid_argument("An error with recoverable meta defintion must provide non-empty field and expected values");
```
cyenyxe: Typo: `defintion`

inc/vcf/error.hpp
```diff
@@ -290,7 +301,10 @@ namespace ebi
                       ErrorFix error_fix = ErrorFix::IRRECOVERABLE_VALUE,
                       const std::string &field = "",
                       const std::string &expected_value = "")
-                : BodySectionError{line, message}, error_fix{error_fix}, field{field}, expected_value{expected_value} {
+                : BodySectionError{line, message, detailed_message},
+                error_fix{error_fix},
+                field{field},
+                expected_value{expected_value} {
                     if (error_fix == ErrorFix::RECOVERABLE_VALUE && expected_value.empty()) {
                         throw std::invalid_argument{"The program had an internal error: An error with recoverable value must provide a non-empty expected value"};
```
cyenyxe: This message seems to be as critical as the one above, but only this one makes it clear enough. I would suggest rephrasing:

> The program had an internal error

as 

> An internal error has occurred

and displaying it in both error classes.

src/vcf/meta_entry.cpp
```diff
@@ -129,18 +129,17 @@ namespace ebi
...
-            if (main_type != DEL
-                && main_type != INS
-                && main_type != DUP
-                && main_type != INV
-                && main_type != CNV
-                && main_type != BND) {
-                throw new MetaSectionError{entry.line, "ALT metadata ID does not begin with DEL/INS/DUP/INV/CNV"};
+            if (!ebi::util::contains(PREDEFINED_INFO_SVTYPES, main_type)) {
+                std::stringstream message;
+                message << "ALT metadata ID containing type and subtype (separated by colon) should have as "
```
cyenyxe: What about this?

> In ALT metadata ID containing colon-separated type and subtypes, the top level type must be one of:

Same in `record.cpp`.

Issue #149, opened 2018-07-19
Incorrect complaint about GP format field in vcf 4.2


kscheffler:
The GP format field is defined as a phred-scaled quantity (i.e. typically larger than 1) in vcf format 4.2 and as a probability between 0 and 1 in vcf format 4.3. However, vcf-validator (version 0.7) reports an error "value does not lie in the interval [0,1]" for vcf 4.2.

tcezard:
Now tracked in EVA-2049: closing

PR #179, opened 2018-12-19
Check  GP in interval [0,1] for v4.3 only.

ssl9ek:
Address issue https://github.com/EBIvariation/vcf-validator/issues/149


jmmut:
thanks for the contribution. our team is out of the office for some days, we'll make sure the fix is correct and complete in some days.

PR #182, opened 2019-03-12
T2D-268 Report missing chromosome if fail to download from ENA

Zhicheng-Liu:
This is to fix an issue when fail to download from ENA, the assembly checker continues to check the reference sequences and report mismatches:

>
>If a contig is not found in ENA, an empty sequence is stored, but it silently fails to check the allele, saying:
>Line 5: Chromosome 1, position 500, reference allele 'C' does not match the reference sequence, expected '' 
>


inc/fasta/fasta.hpp
```diff
@@ -34,8 +34,28 @@ namespace ebi
...
+         * @param contig - the name of the contig to extract the sequence from
+         * @param pos - the starting offset position
+         * @param length - the number of base pairs to be extracted
+         * @return the sequence string, empty if nothing can be extracted.
+         */
         virtual std::string sequence(const std::string& contig, const size_t pos, const size_t length) = 0;
-        virtual size_t count(const std::string& contig) const = 0;
+
+        /**
+         * Get the occurrences of sequences for a contig
```
cyenyxe: I think this may be a bit clearer:

```suggestion
         * Get the number of times a contig appears in a FASTA.
```

src/vcf/assembly_checker.cpp
```diff
@@ -121,7 +121,7 @@ namespace ebi
           if (rc.get()) {
               fetch_contig_on_demand = true;
           }
-
+          std::unordered_set<std::string> downloadedContigs;
```
cyenyxe: This should be named `downloaded_contigs` in order to follow our naming conventions.

src/vcf/assembly_checker.cpp
```diff
@@ -150,12 +150,17 @@ namespace ebi
...
                   contig_name = found_synonyms[0];
-              } else if (!fetch_contig_on_demand) {
-                  if (fasta->count(contig_name) == 0) {
-                      report_missing_chromosome(line_num, record_core, outputs);
-                      is_valid = false;
-                      continue;
-                  }
+              }
+
+              if (fetch_contig_on_demand && downloadedContigs.find(contig_name)==downloadedContigs.cend()) {
```
cyenyxe: Add whitespaces around the comparison operator. Applies to other conditions below.

test/fasta/fasta_test.cpp
```diff
@@ -0,0 +1,135 @@
...
+
+      SECTION("Check if contigs exist, whole genome")
+      {
+          std::unique_ptr<ebi::vcf::fasta::IFasta> fasta(
+                  new ebi::vcf::fasta::FileBasedFasta(WHOLE_GENOME_FASTA, WHOLE_GENOME_FASTA_INDEX));
+
+          CHECK(fasta->count("I") == 1);
+          CHECK(fasta->count("II") == 1);
+          CHECK(fasta->count("III") == 1);
+          CHECK(fasta->count("Unexisted") == 0);
```
cyenyxe: Not critical, but for language correctness:

```suggestion
          CHECK(fasta->count("Non-existing") == 0);
```

Applies to others below.

src/fasta/fasta.cpp
```diff
@@ -88,6 +102,7 @@ class ebi::vcf::fasta::ContigFromENA {
   std::string read(const size_t pos, const size_t length)
   {
       if (!fasta_input.get()) {
+          fasta_output->flush();
```
jmmut: I'm wondering if all this details could be avoided if we simplified the design of these classes. Is there any reason not to make `fasta_input` and `fasta_output` regular fstreams instead of pointers? that way they would be initialized in the constructor and destroyed automatically in the destructor.

src/vcf/assembly_checker.cpp
jmmut: why do we need this set if there are also methods in IFasta to ask if the contig is available, and download it automatically if not?

src/vcf/assembly_checker.cpp
```diff
@@ -121,7 +121,7 @@ namespace ebi
           if (rc.get()) {
```
jmmut: taking another look at this, why does the user (assembly_checker.cpp) need to know the IFasta implementation? if the main created one or other implementation based on the parameters, the implementations should act differently so that there shouldn't be checks here on `fetch_contig_on_demand`, right?

src/fasta/fasta.cpp
Zhicheng-Liu: This one was not because of using pointer for `fasta_output`. It was caused by separating output and input streams for the same underlying temporary file. I wanted to use a same `fstream` object for writting the downloaded fasta to and reading the sequence from. But at that point, I could not get it working.
I have revisited this issue and now found a proper way to open a `fstream` for both input and output, as changed in d599e3b. Also refactor other places that use smart pointer unnecessarily.
Thank you very much for pointing this out. Please let me know anything further I can do.

src/vcf/assembly_checker.cpp
Zhicheng-Liu: Indeed the fasta utility class provides `count` method which can be used to check if an atempt for downloading a contig has been made. Removed in d599e3b.

src/vcf/assembly_checker.cpp
Zhicheng-Liu: Yes, this was ugly. Partly was due to previous implementation which requires passing a boolean as an argument in long chain of function calls. This has now been refactored in d599e3b. Hope this looks cleaner.

jmmut:
Also, not strictly inside this ticket, but if a position in the VCF is greater than the sequence length in the fasta, the same message appears: 

> Line 3: Chromosome II, position 2000, reference allele 'A' does not match the reference sequence, expected ''

unlike in the bug that this PR solves, this time it's always correct that it's a mismatch, but the message is misleading in my opinion.

inc/fasta/fasta.hpp
```diff
@@ -34,8 +33,28 @@ namespace ebi
...
+         */
         virtual std::string sequence(const std::string& contig, const size_t pos, const size_t length) = 0;
-        virtual size_t count(const std::string& contig) const = 0;
+
+        /**
+         * Get the number of times a contig appears in a FASTA.
+         * @param contig - the name of the contig
+         * @return the number of times a contig appears in a FASTA. 0 if the contig is not found.
+         */
+        virtual size_t count(const std::string &contig) const = 0;
```
jmmut: does it really make sense to call this `count`? having the other method for the sequence length, this becomes more and more obscure. I would call it `bool is_present` or `bool is_contig_present`. I guess this is called `count` after bioio::FastaIndex::count which is actually std::unordered_map::count which can only return 0 or 1.

inc/fasta/fasta.hpp
```diff
@@ -34,8 +33,28 @@ namespace ebi
...
+
+        /**
+         * Get the length of the sequence for a contig
+         * @param contig - the name of the contig
+         * @return the length of the sequence for the contig. 0 if the contig is not found.
+         */
+        virtual size_t len(const std::string &contig) const = 0;
```
jmmut: unnecessary abbreviation, in my opinion. Actually, if you think it's useful to say in the documentation that it's the *sequence* length, I would call the method `sequence_length`.

inc/fasta/fasta.hpp
cyenyxe: And if it can return a greater number, then `sequence_num_occurrences` would still be clearer. I suggest the `sequence` prefix for consistency with the rest of methods or suggestions about them.

PR #183, opened 2019-04-01
T2D-268 Add support for on-the-fly fasta indexing

Zhicheng-Liu:
With this change, when using assembly checker with missing local fasta index file, the assembly checker will create the index for input fasta on-the-fly and complete the assembly checking as normal, instead of throwing an error as before.

Depends on the changes in https://github.com/EBIvariation/vcf-validator/pull/182.


src/fasta/faidx.cpp
```diff
@@ -0,0 +1,199 @@
...
+    std::string name;     // Name of this contig
+    uint64_t seq_length;  // Total length of this contig, in bases
+    uint64_t seq_offset;  // Offset in the FASTA file of this contig's first base
+    uint32_t line_length; // The number of bases on each line
+    uint32_t line_bytes;  // The number of bytes in each line, including the newline
+};
+
+class FastaIndexerState;
+
+class FastaIndexer {
```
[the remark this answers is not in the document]
Zhicheng-Liu: Well, it may appear to be like a missile, in my eyes it is a swiss army knife. For parsing a fasta file, yes, either way is clear and simple. I just felt a bit more comfortable with separating out each states.
I have put in another change in aa50e83 to address the (pseudo-)recursive calls. I didn't like them either but I don't know why I did not figure out the iterative way. Maybe it was because the fact that most (if not all) fasta files do not contain a lot of contigs made me relax with the call stacks. Anyway, with the new change this issue should have been resolved.

src/fasta/faidx.cpp
```diff
@@ -0,0 +1,190 @@
...
+void
+NewContigState::process(FastaIndexer *fastaIndexer, std::istream & input) {
+    fastaIndexer->fastaIndex.push_back({"", 0, 0, 0, 0});
+    fastaIndexer->set_state(new ContigNameState());
+}
+
+void
+ContigNameState::process(FastaIndexer *fastaIndexer, std::istream & input) {
+    char c;
+    ContigIndex & contigIndex = fastaIndexer->fastaIndex[fastaIndexer->fastaIndex.size()-1];
```
jmmut: I would have used 
```suggestion
    ContigIndex & contigIndex = fastaIndexer->fastaIndex.back();
```
but the current version is ok. longer but clearer.

Zhicheng-Liu: I will stick with the original implementation as that's clearer to me.

src/fasta/fasta.cpp
```diff
@@ -36,9 +37,19 @@ ebi::vcf::fasta::FileBasedFasta::FileBasedFasta(const std::string& fasta_path, c
...
+        std::ifstream fs;
+        ebi::util::open_file(fs, fasta_index_path, std::ifstream::binary);
+        fasta_index = bioio::read_fasta_index(fs);
+    } catch (const std::runtime_error& error) {
+        // Fasta index file does not exist. Create index on-the-fly.
+        BOOST_LOG_TRIVIAL(info) << "Creating index from input FASTA file...";
+        std::stringstream faidx_index;
+        ebi::vcf::faidx::index_fasta(fasta_input, faidx_index);
+        fasta_input.clear();
+        fasta_input.seekg(0, std::ios::beg);
```
cyenyxe: Given that these 2 actions need to be happen for the index to be read properly after creation, maybe it would make sense to reset the stream at the end of the `index_fasta` method? That way it will always be created and leave everything ready.

Zhicheng-Liu: In `index_fasta`, the reference is `std::ostream` type so you can only set the pointer for `put` position through `seekp` call. That would have no effect on the `get` position pointer if these two pointers are different as in most (if not all) implementations.

However I am told that whether to use two separate pointers for `get` and `put` is implementation dependent. In the case in which the same pointer is used for `get` and `put` these two lines would be needed here. In most cases in which two separate pointers are used, these two lines are not necessary, but I would like to leave them here to make the intention clear.

PR #185, opened 2019-05-07
EVA-1474 Validator fails on ubuntu 18.04 (only update README)

jmmut:
This is not a fix for the bug, only a notice in the README so that users know that we are aware of the problem and a possible workaround

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,6 +17,14 @@ We recommend using the [latest release](https://github.com/EBIvariation/vcf-vali
 
 ## Run
 
+#### Note: Core dumped in Ubuntu 18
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -17,6 +17,14 @@ We recommend using the [latest release](https://github.com/EBIvariation/vcf-vali
 
 ## Run
 
+#### Note: Core dumped in Ubuntu 18
+We recently noticed that the current version 0.9.1 fails to execute in Ubuntu 18 with the next message:
```
cyenyxe: To avoid having to update it until we fix it:

```suggestion
We recently noticed that the current version fails to execute in Ubuntu 18 with the next message:
```

PR #187, opened 2019-05-21
Clean assembly_checker

iyangar:
moving internal functions to source file

### DIFF
--- a/src/vcf/assembly_checker.cpp
+++ b/src/vcf/assembly_checker.cpp
@@ -22,21 +22,27 @@ namespace ebi
   {
     namespace assembly_checker
     {
-      bool process_vcf_records(std::istream & vcf_input,
-                               std::shared_ptr<ebi::vcf::fasta::IFasta> & fasta,
-                               const std::string & assembly_report,
-                               std::vector<std::unique_ptr<ebi::vcf::AssemblyCheckReportWriter>> & outputs,
-                               bool use_fasta_from_ena);
-
-      std::string get_reference_accession(const std::string& reference_tagged_line);
-      std::string get_contig_accession(const std::string& contig_tagged_line);
+      size_t const DEFAULT_LINE_BUFFER_SIZE = 64 * 1024;
 
       bool process_vcf_ref(std::istream & vcf_input,
                            const std::string & fasta,
                            const std::string & assembly_report,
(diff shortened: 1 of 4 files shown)


src/vcf/assembly_checker.cpp
```diff
@@ -31,6 +31,31 @@ namespace ebi
       std::string get_reference_accession(const std::string& reference_tagged_line);
       std::string get_contig_accession(const std::string& contig_tagged_line);
 
+      bool process_vcf_ref(std::istream & vcf_input,
```
jmmut: we usually order the functions so that the functions appear right after they are used.
if the code is 
```
function a {
    b()
    c()
} 
function b {
    called_from_b()
}
```
we sort them (prototypes too) like this:
```
a()
b()
called_from_b()
c()
```
now that you are cleaning the prototypes in this file, can you order them too?

iyangar:
https://github.com/EBIvariation/vcf-validator/pull/187#pullrequestreview-240097523
Sure, i will move. `ebi::vcf::default_line_buffer_size` and several other constants are in small? Let me know if i have to change here

jmmut:
I can't see any other global constants in lower case, can you give an example? anyway I would only change default_line_buffer_size for now because it's better to reduce the number of files changed in each PR.

iyangar:
ebi::vcf::default_line_buffer_size is used in `compressed_file_test.cpp`. I will change to upper case and leave it in the header file.

src/vcf/assembly_checker.cpp
iyangar: Here is skeleton of functions with ordering. If my interpretation is correct i will change both declarations and definitions in this order.
---------------------------
check_vcf_ref()
 - process_vcf_ref()
 
process_vcf_ref()
 - get_reference_accession()
 - get_contig_accession()
 - process_vcf_records()
 
get_reference_accession()

get_contig_accession()

process_vcf_records()
 - build_record_core()
 - report_telomere_position()
 - get_matching_synonyms_list()
 - report_missing_chromosome()

build_record_core()

report_telomere_position()

get_matching_synonyms_list()
 - report_missing_chromosome()
 - report_multiple_synonym_match()

report_missing_chromosome()
 
report_multiple_synonym_match()

Issue #190, opened 2019-08-05
Intermittent 'number of samples' error


EvanTheB:
```4561:Error: The number of samples must match those listed in the header line. This occurs 1 time(s), first time in line 1070259.```

I am getting this error intermittently, on the same file. Sometimes no error. Sometimes an error (on a different line). 

VCF is gzipped.

Version is vcf_validator_linux from this github release page:
```
$ vcf_validator --version
vcf_validator version 0.9
```

I have visually inspected, and cannot identify a reason for the error. 

This script gives me only 1 value, the expected number of samples:
```
zcat ../reheader/MGRB.phase2.SNPtier12.match.vqsr.minrep.WGStier12.unrelated.nocancer.over70.21.vcf.gz | awk '/^[^#]/{print NF}' | uniq
```

Any ideas or further tests I can perform?

EDIT:
As one further test, I ran vcf_validator under valgrind, there were thousands of errors.

jmmut:
Thanks for the report. We have seen this a couple times before with big files, but couldn't fix it because we couldn't reproduce it properly. We only have a vague idea where the problem could be. Maybe the gzip decompression, maybe the parser, maybe both.

How big is your file? does it really have 1070259 lines or more? How did you run the validator? `vcf_validator -i your_file.vcf.gz`? or `zcat your_file.vcf.gz | vcf_validator`?

Regarding the valgrind errors, it's possible that they are related to this issue, although most of them are errors in pthread and boost which are outside of our control. (Some time ago we did go through valgrind errors and fixed most of them, you can see that "All heap blocks were freed -- no leaks are possible").

Is your VCF open access data? I'm thinking if you could send your file to us. I can provide an FTP where you can upload it if it's big.

If nothing else works, you could always split your file and copy the header to each part. Sorry for the inconvenience of this bug, there is always something more urgent to fix.

EvanTheB:
The file is 100Gb, it is not open access sorry.

I am seeing this pretty consistently on files as small as 100Mb though. 

The valgrind errors I am seeing are branches based on uninitialised memory. I would suggest that if your libraries are invoking undefined behaviour, you stop using them.

jmmut:
Then, what about sharing one of those 100Mb VCFs?  Any chance that you could anonymise it? Or at least describe in general terms the features of those VCFs? Like number of samples, fields used in INFO and FORMAT, if it's mostly SNPs or mostly SVs, etc, I don't really know what we should look for.

[...]

Issue #184, opened 2019-04-08
Error with loadlocale: _nl_intern_locale_data


dariober:
Hi- with vcf_validator v0.9.1 on Ubuntu 18.04 I get the error:

```
./vcf_validator -i data/common.sample.vcf.gz
[info] Reading from input file...
vcf_validator: loadlocale.c:129: _nl_intern_locale_data: Assertion `cnt < (sizeof (_nl_value_type_LC_TIME) / sizeof (_nl_value_type_LC_TIME[0]))' failed.
Aborted (core dumped)
```

After some googling I found I can work around it by first executing  `export LC_ALL=C` but it would be nice to have a more permanent solution...

jmmut:
Hi, yes, you are right, we noticed this some days ago and reached to the same conclusion. However, we didn't have time to fix this yet. I can put a short notice in the readme for now. We will update and close this issue when it's fixed. Thanks.

jmmut:
This is from april... 6 months already... Be sure I haven't forgotten about this. I'll try a bit harder to convince my bosses to let me (or someone else) spend some time on this. If there's a PR from anyone I'll probably take a look in my personal spare time.

jmmut:
fixed in the latest release: v0.9.3: https://github.com/EBIvariation/vcf-validator/releases/tag/v0.9.3

PR #197, opened 2020-01-28
EVA-1637 vcf validator and asm checker can't uncompress bgzip files properly

jmmut:
History of this issue:
- I generated a **b**gzipped VCF that passes validation if it's decompressed before calling the validator (`zcat file.vcf.gz | vcf_validator`), but doesn't pass validation if the validator itself decompresses the file (`vcf_validator -i file.vcf.gz`). If the file is decompressed and gzipped, it also works. It seems that depending on the block size, the program didn't read properly all lines.
- A test was added to process this file (test/input_files/v4.3/compressed_files/readable/passed/boost_55_can_not_decompress_last_line_of_this_bgzipped.vcf.gz).
- Confirmed with the bgzip people at 
https://github.com/samtools/htslib/issues/953 that bgzip should be correctly read by any gzip reader. We found out that newer versions of boost decompress the offending file correctly.
- Changed the build in linux to download and compile Boost, instead of installing from distro packages.
  - [x] Changed the install_dependencies.sh script for manual installation.
  - [x] Changed the travis linux build.
  - [ ] Changed the docker build. NOT DONE: it seems this failed even in master, so it's out of the scope of this ticket. The docker build is removed for now.
  - [x] Changed README to reflect the new (and easier) installation steps.
- The OSX validator passes the test without modification, so the build was not changed for OSX.
- Decompression for Windows has never worked and is low priority, so no changes there.
- The binaries from travis for the release will be compiled with gcc 5 (the xenial default). The travis build for gcc 4.8 and gcc 6 were removed because they were not working, possibly due to ABI incompatibilities with ODB. Maybe compiling all dependencies with the same ABI fixes this, but that's out of scope of this ticket.
- The locale problem https://github.com/EBIvariation/vcf-validator/issues/184 prevents from building in Ubuntu 18, but the precompiled binary (built in Ubuntu 16) can be run without problems in Ubuntu 18, so that issue is half done.

The cause of https://github.com/EBIvariation/vcf-validator/issues/190 wasn't identified, but this PR maybe fixes it, as the validator now reads correctly bgzip files.


install_dependencies.sh
```diff
@@ -68,27 +71,39 @@ mkdir -p $dependencies_dir && cd $dependencies_dir
 
 echo "installing libodb"
 wget http://codesynthesis.com/download/odb/2.4/libodb-2.4.0.tar.bz2 -O /tmp/libodb.tar.bz2
-tar jxvf /tmp/libodb.tar.bz2
-cd libodb-2.4.0 && ./configure && make && cd ..
+tar jxf /tmp/libodb.tar.bz2
+cd libodb-2.4.0 && ./configure && make
+RESULT=$?
+INSTALL_FAILED=$(( $INSTALL_FAILED + $RESULT ))
```
sundarvenkata-EBI: Will a bitwise-or be better here? I am guessing bash won't return negative exit codes, so we are covered here, right?

install_dependencies.sh
```diff
@@ -19,6 +19,7 @@ it installs the given dependencies:
   - curl library                            curl-7.62.0
   - openssl library                         openssl-1.1.1a
   - c-ares library                          c-ares-1.15.0
+  - boost library                           boost-1.72.0
 
 for linux:
 ./install_dependencies.sh linux
```
sundarvenkata-EBI: Are elevated permissions needed for running this script (make outputs etc.,)? In that case, should we call out this script to be executed as a privileged user?

sundarvenkata-EBI: Never mind, looks like the make install commands seem to target the local directory anyway...

install_dependencies.sh
jmmut: please suggest a syntax because I can't make the "or" work.

jmmut: actually, I realised there's a `set -e` at the top that will make the script exit on the first failure, so I will leave only the "everything ok" at the end, as the "install failed" message will never show

PR #198, opened 2020-03-30
EVA-1199 Validate genotypes and frequencies are present

andresfsilva:
The flag `--require-evidence` can be passed to the vcf validator to indicate it should check the VCF has either Genotypes or Allele Frequencies


README.md
```diff
@@ -45,6 +45,8 @@ Different types of validation reports can be written with the `-r` / `--report`
 
 Each report is written into its own file and it is named after the input file, followed by a timestamp. The default output directory is the same as the input file's if provided using `-i`, or the current directory if using the standard input; it can be changed with the `-o` / `--outdir` option.
 
+A flag for validating evidence can be passed with `--require-evidence`. This will validate the VCF includes either Genotypes or Allele Frequencies, by default this flag will be false.
```
sundarvenkata-EBI, line 48:
```suggestion
A flag for validating evidence can be passed with `--require-evidence`. This will validate whether the VCF includes either Genotypes or Allele Frequencies. This flag will be false by default.
```

inc/vcf/validator_detail_v41.hpp
```diff
@@ -11037,7 +11039,9 @@ case 533:
 #line 60 "src/vcf/vcf.ragel"
 	{
         ErrorPolicy::handle_error(*this, new FileformatError{n_lines});
-        p--; {goto st519;}
+        p--; {       if ( p == pe )
+               goto _test_eof519;
+goto st519;}
```
sundarvenkata-EBI, line 11044: Is this auto-generated code?

andresfsilva: yes, this is done by ragel

src/vcf/parsing_state.cpp
```diff
@@ -87,5 +91,31 @@ namespace ebi
...
         defined_metadata.emplace(meta_type, id);
     }
+
+    void ParsingState::validate_additional_checks()
+    {
+        if (additionalChecks.checkEvidence) {
+            if (genotypes_present() || allele_frequencies_present()) {
+                return;
+            }
+            throw new InfoBodyError{record->line, "Genotypes or Allele frequencies are required"};
```
jmmut, line 101: if there are no genotypes, that's not exactly a INFO body error. I would use a BodySectionError. Also, please explain better to the user what exact fields are supposed to be filled. 

I would just use the basic message, making the string a bit longer (but concise!). It doesn't apply to this error, but for other errors we use the detailed message, and the difference between the basic message and the detailed message, is that we group by the basic message (exact string), and the detailed message has concrete data about the error (that will change across lines), e.g. `GL was "0,1", but expected 3 values`.

test/vcf/parser_v43_test.cpp
```diff
@@ -81,5 +81,42 @@ namespace ebi
...
+
+      SECTION("With genotypes or frequencies") {
+          auto folder = boost::filesystem::path("test/input_files/v4.3/evidence/passed");
+          std::vector<boost::filesystem::path> v;
+          copy(boost::filesystem::directory_iterator(folder), boost::filesystem::directory_iterator(), back_inserter(v));
+
+          for (auto path : v)
+          {
+              SECTION(path.string() + " Validate evidence")
+              {
```
jmmut, line 111: I don't care that much about which style we use, but at least make it locally consistent.
```suggestion
          for (auto path : v) {
              SECTION(path.string() + " Validate evidence") {
```

src/vcf/parsing_state.cpp
```diff
@@ -87,5 +91,31 @@ namespace ebi
...
+        }
+    }
+
+    bool ParsingState::genotypes_present() {
```
jmmut, line 105:
```suggestion
    bool ParsingState::genotypes_present()
    {
```

src/vcf/parsing_state.cpp
```diff
@@ -87,5 +91,31 @@ namespace ebi
...
+        auto format = record->format;
+        return !format.empty() && util::contains(format, GT);
+    }
+
+    bool ParsingState::allele_frequencies_present()
+    {
+        auto info = record->info;
+        for (auto &in : info) {
+            if (in.first == vcf::AF) {
```
jmmut, line 114: `auto info = record->info` is making a copy. Either use a reference `auto &info = ...` or just use it inlined. In this case I think the variable doesn't add clarity.
```suggestion
        for (const auto &info_key_value : record->info) {
            if (info_key_value.first == vcf::AF) {
```
also, put `const` in every read-only reference.

src/vcf/parsing_state.cpp
```diff
@@ -87,5 +91,31 @@ namespace ebi
     {
         defined_metadata.emplace(meta_type, id);
     }
+
+    void ParsingState::validate_additional_checks()
+    {
+        if (additionalChecks.checkEvidence) {
+            if (genotypes_present() || allele_frequencies_present()) {
```
jmmut, line 98: I forgot: it's ok too if the VCF only has allele counts (AC and AN). The frequency can be calculated from that.

src/vcf/parsing_state.cpp
```diff
@@ -87,5 +91,55 @@ namespace ebi
...
+    }
+
+    bool ParsingState::allele_count_present()
+    {
+        bool allele_count_present = false;
+        bool allele_total_number_present = false;
+        for (const auto &info_key_value : record->info) {
+            if (info_key_value.first == vcf::AC) {
+                if (allele_total_number_present) {
+                    return true;
```
jmmut, line 131: funny algorithm!

PR #199, opened 2020-04-01
EVA-1899 - Fix GT handling for CN variants + fix stoi handling in other places

sundarvenkata-EBI:
[Fix high-level Integer checking](https://github.com/EBIvariation/vcf-validator/pull/199/files#diff-b7ed7bfe2cd51f747bc5f560a4647ee3R730) to ensure that strings like "1A" parsed by stoi to a number like "1" are not accepted as it is. 

The other stoi code updates are purely from a "defensive coding" perspective because there are other checks like [this](https://github.com/EBIvariation/vcf-validator/pull/199/files#diff-b7ed7bfe2cd51f747bc5f560a4647ee3R420) which are not absolutely necessary because such checks are superseded by the high-level integer checking described above.

### DIFF
--- a/src/vcf/record.cpp
+++ b/src/vcf/record.cpp
@@ -417,16 +417,16 @@ namespace ebi
                 } else {
                     std::string first_field = alternate_alleles[i].substr(0, 4);
                     if (first_field == "<" + INS || first_field == "<" + DUP) {
-			size_t scanned_value_length;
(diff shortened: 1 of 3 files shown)


sundarvenkata-EBI:
I will re-run the failed checks on Mac OS after some time has passed so as to not aggravate the OpenSSL DDOS mechanisms...

PR #201, opened 2020-04-26
EVA-1950 VCF validator fails to detect some duplicate variants

jmmut:


jmmut:
The travis build is ok. I reordered the commits and github doesn't show them in branch order. the travis build passed on the latest commit.

inc/vcf/record_cache.hpp
```diff
@@ -105,6 +105,25 @@ namespace ebi
             list_duplicates.clear();
             list_symbolic_duplicates.clear();
 
+            if (current_chromosome != record.chromosome) {
+                // EVA-1950: the cache removes "lexicographically lower" chromosomes when the cache is full, but with
+                // that sorting, "chr10" is lower than "chr2", so new variants in "chr10" get deleted to keep the ones
+                // in "chr2", and duplicates in "chr10" will go undetected. Possible solutions (better first):
+                // 1. The chosen simple solution is to keep in the cache variants from only one chromosome.
```
tcezard, line 112: How will this solution scale with VCF with millions of contigs?

inc/vcf/record_cache.hpp
```diff
@@ -105,6 +105,25 @@ namespace ebi
...
+                // 2. Another solution could be using the line number (Record::line) as sorting criteria, but that can
+                // get very surprising and unintuitive if used for other purposes outside of this class, so a custom
+                // comparator could be provided to the cache multiset constructor.
+                // 3. A third solution could be making a Least Recently Used cache, but will require using esoteric data
+                // structures and/or dependencies, or keeping several simpler structures in sync, like a map and a
+                // queue that point to each other's contents.
+                // 4. A poor solution could be changing the ordering to "numerically lower", but that will be very tricky
+                // to get right with weird chrs like (versioned) accessions. e.g.: NC_01.1 is a chr, which is likely to
+                // appear before a contig GK_01.1, and we would need a sorting method that evaluates NC_01.1 < GK_01.1.
```
tcezard, line 121: I appreciate the comments that help describe your rational but I'm not sure this is the best format to document the chosen solution, most of the text refers to solution that have not been selected and I would move this to either comments in the PR or comments in the JIRA tickets.

inc/vcf/record_cache.hpp
jmmut: it's unaffected by the number of contigs. This solution keeps the variants of one chromosome *at a time*, so when the next one comes, the cache is cleared and the next chr can fill the cache. I doesn't matter how many chrs/contigs there are. 
The only hypothetical case that this cache is not enough by default is if there are  999 variants between 2 duplicate variants. In that case the 2 duplicate variants would pass undetected. I can give some artificial examples that would trigger that, but they look quite unrealistic.

inc/vcf/record_cache.hpp
jmmut: IMO the comments that explain why some solution was not chosen are the best kind of comments. When someone looks at this and doesn't understand why a proper LRU cache was not used instead of this cheap solution, it will be difficult to find the reasoning if it's in the PR on in the (private) JIRA. I'm going to reword it a bit, but if you still think it's better to move it out, I will do that.

tcezard: Ok that's fine

PR #202, opened 2020-05-18
Detected compression format now a debug message instead of warning

cyenyxe:
As mentioned in https://github.com/EBIvariation/vcf-validator/pull/117#issuecomment-628762553, compressed files are now read correctly and there is no point in raising that as a warning, but more as a debug message.


jmmut:
thanks! :)

PR #208, opened 2020-10-20
Parse assigned molecule in assembly report

tcezard:
Allow the Assembly checker to use assigned molecule as a source of name.


tcezard:
This won't work: closing for now

PR #210, opened 2021-01-15
EVA-2188 - Exit assembly checker when no Genbank synonyms are found for a contig

sundarvenkata-EBI:


tcezard:
I'm not sure adding this functionality in the assembly checker is actually what we want. 
I think I agree that we want to check that the contigs all have a INSDC accession but doing it in the assembly checker seems wrong since its purpose is exclusively to check the reference base.
Here you're validating the assembly report instead of the VCF.

sundarvenkata-EBI:
> Here you're validating the assembly report instead of the VCF.
This is a fair point. But reporting the presence of non-Genbank contigs requires parsing the assembly report as well as scanning the VCF for offending contigs. Currently, the technical machinery for both are available in assembly checker. So this was the more pragmatic choice. As we discussed in DEV meeting today, I will 1) add a command-line flag for the user to explicitly state that a VCF is meant for submission to EVA and 2)  make sure that this check is performed only when the flag is set.

inc/assembly_report/assembly_report.hpp
```diff
@@ -77,6 +78,19 @@ namespace ebi
...
+         * Check if a contig synonym is in a genbank accession format
+         * @see <a href="https://www.ncbi.nlm.nih.gov/Sequin/acc.html">Genbank accession format</a>
+        */
+        static bool is_a_genbank_accession(const std::basic_string<char>& synonym) {
+            static const boost::regex genbank_accession("[A-Z][0-9]{5}\.[0-9]+|"
+                                                        "[A-Z]{2}[0-9]{6}\.[0-9]+|"
+                                                        "[A-Z]{2}[0-9]{8}\.[0-9]+|"
+                                                        "[A-Z]{4}[0-9]{8}\.[0-9]+|[A-Z]{6}[0-9]{9}\.[0-9]+" //WGS sequences
+            );
+            return boost::regex_match(synonym, genbank_accession);
```
tcezard, line 91: What is the purpose of the pipe `|` at the end of the regex ?

sundarvenkata-EBI: It is an OR delimiter: the accession can fit **_any_** of the said patterns.

tcezard: Ah ok.  I thought, it was a list of string but they're all concatenated obviously.

inc/assembly_report/assembly_report.hpp
```diff
@@ -166,6 +180,12 @@ namespace ebi
         {
             std::vector<int> synonym_indices{0, 4, 6, 9};
             ContigSynonyms contig_synonyms;
+            //For unequal entries in the assembly report, discard the Genbank contig since it is not useful anyway
+            //ex: Pltd	assembled-molecule	Pltd	Chloroplast	X19501.1    <>	NC_001320.1	non-nuclear	134525	na
+            if (boost::trim_copy(synonyms[5]) == "<>") {
+                contig_synonyms.add_synonym(synonyms[0]);
+                return contig_synonyms;
```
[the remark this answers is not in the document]
sundarvenkata-EBI: Fair point. I looked up the NCBI assembly help reference and they do say that non-equivalent contigs are marked as "na". I have undone this change!

andresfsilva (approved):

tcezard (approved):

PR #216, opened 2022-02-17
I know this is probably obvious to you, but not me

CholoTook:
Any info you can give the user to save them confusion is a plus IMHO.


tcezard (approved):
Thank you. It never hurts to be precise.

PR #217, opened 2022-04-28
EVA-2844 - Port linux and macOS build to Github actions

apriltuesday:
Github actions behaviour is as follows:
* On pull request and push to master, build and run tests
* On tag, also create a draft release and upload artifacts

There is a sample successful run [here](https://github.com/apriltuesday/vcf-validator/actions/runs/2240743046) and release [here](https://github.com/apriltuesday/vcf-validator/releases/tag/test-release) (published so viewable by others).


install_dependencies.sh
```diff
@@ -125,13 +125,34 @@ cd c-ares-1.15.0
 make && make install
 cd ..
 
+if [[ "$OS_NAME" == "osx" ]]
+then
+  echo "installing nghttp2"
+  mkdir nghttp2
+  wget https://github.com/nghttp2/nghttp2/releases/download/v1.47.0/nghttp2-1.47.0.tar.gz -O ./nghttp2-1.47.0.tar.gz
```
sundarvenkata-EBI, line 132: Suggest parametrizing version numbers to a variable so that we can more easily manager version increments later

sundarvenkata-EBI (approved):

tcezard (approved):

install_dependencies.sh
apriltuesday: This is a good idea but not trivial for all the dependencies in the script...  For this ticket I think I'll skip it, but I did update the help message so all the libraries and versions are at least listed in one central place.

PR #224, opened 2023-07-13
EVA-3338 - Missing data is valid even if multiple values are expected

tcezard:


sundarvenkata-EBI:
@tcezard Wondering if we should add tests for the specific use-case that we are fixing here...

sundarvenkata-EBI (approved):

apriltuesday (approved):

PR #233, opened 2024-01-19
fix for build failure due to minmac version

vasudeva8:
#230 had a change specific for Mac but not guarded with os check and caused failure in linux build as well.
Now the change is guarded appropriately.


tcezard (approved):
That makes sense.

apriltuesday (approved):

PR #237, opened 2024-02-05
GA4GHTT-226 - updates for Mac build - 2

vasudeva8:
Changes to have Mac build - part 2.

Uses Mac version 12 instead of 11 for build.
Update of home-brew is disabled as build fails at times on environment setup.
zstd path added along with icu on library path export.
Enabled Mac build configuration in build.yml and added to release job as well.
Mac build uses libcurl provided by OS instead of building from source, as the build fails due to home-brew python installation issues.

### DIFF
--- a/install_dependencies.sh
+++ b/install_dependencies.sh
@@ -20,14 +20,12 @@ it installs the given dependencies:
   - odb compiler                            odb-2.4.0
   - odb common runtime library              libodb-2.4.0
   - odb sqlite runtime library              libodb-sqlite-2.4.0
-  - bzip library                            bzip2-1.0.6
-  - zlib library                            zlib-1.2.11
-  - curl library                            curl-7.62.0
-  - openssl library                         openssl-1.1.1w
-  - c-ares library                          c-ares-1.15.0
-  - boost library                           boost-1.72.0
-  - nghttp2 library (osx only)              nghttp2-1.47.0
-  - brotli library (osx only)               brotli-1.1.0
+  - bzip library (linux only)               bzip2-1.0.6
+  - zlib library (linux only)               zlib-1.2.11
+  - curl library (linux only)               curl-7.62.0
+  - openssl library (linux only)            openssl-1.1.1w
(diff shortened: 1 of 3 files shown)


install_dependencies.sh
```diff
@@ -106,98 +104,56 @@ fi
...
-  wget https://github.com/nghttp2/nghttp2/releases/download/v1.47.0/nghttp2-1.47.0.tar.gz -O ./nghttp2-1.47.0.tar.gz
-  tar xzf ./nghttp2-1.47.0.tar.gz
-  cd nghttp2-1.47.0
-  ./configure --prefix=$dependencies_dir_abs_path/nghttp2
-  make && make install
+  dependencies_dir_abs_path=`pwd`
+
+  echo "installing openssl"
+  mkdir openssl
+  #using latest openssl which supports mac arm as well
```
tcezard, line 130:
```suggestion
  #using openssl v1.1.1w which supports mac arm as well
```

install_dependencies.sh
```diff
@@ -106,98 +104,56 @@ fi
...
-make && make install_sw
-cd ..
+if [[ "$OS_NAME" != "osx" ]]
+then
+  #linux 
+  echo "installing libbz2"
+  # This is commented till the bzip.org site is recovered.
+  # wget http://www.bzip.org/1.0.6/bzip2-1.0.6.tar.gz -O ./libbz2.tar.gz
+  # tar zxf ./libbz2.tar.gz
+  # Till then we can trust ubuntu archives.
```
tcezard, line 114: This comment was added [6 years ago](https://github.com/EBIvariation/vcf-validator/commit/74949d848c85475382f18d4743f1b295dfdc1dbd). I think we can remove it since it's unlikely that their wordpress website  would come back anytime soon.
```suggestion
```

tcezard (approved):
Changes looks great. I left a couple of small suggestions

apriltuesday (approved):
Great work, thanks!

PR #240, opened 2024-03-13
GA4GHTT-232 - vcf4.4 initial changes

vasudeva8:
Initial changes to support VCF v4.4.

Made v4.4 ragel file, using v4.3 ragel with version change alone, that it parses files conforming to v4.3 with v4.4 as version gets validated.
v4.4 specific changes are to be identified and updated.

Test binaries and files updated (there were a few failed case data files which couldn't be opened / updated are left as such; also files which were not matching to v4.3 already are left as such).

test suit, v4.4 specific test binary and other versions works without failures. assembler, debugulator and validator works but further tests to be done.


docs/developer-guide.md
```diff
@@ -41,6 +41,10 @@ executable binaries:
 
 ![architecture image](img/VCF-validator-architecture.png)
 
+In case of dependent libraries being installed in non-standard paths,
+update LIBRARY_PATH as shown below
+  'export LIBRARY_PATH=$LIBRARY_PATH:<path to required library's libpath>'
+
```
tcezard, line 47: I'm not sure in which case this applies. 
I've not seen this being used anywhere in the `CMakeList.txt` and it didn't alleviate my compilation issues.

tcezard (approved):
This PR looks good. I've updated the target branch to dev/vcf4.4

docs/developer-guide.md
vasudeva8: This could be useful when dependent libraries, expected to be available through OS,  were installed in paths which are not part of library search. During Mac OS builds, I used to get link issues with zstd and icu libraries. They were present inside home-brew folders which were not part of any lib search environment variables.

Tried different ways to resolve them, like modifying CMake file, updating home-brew/cmake variables etc., and later noticed this from build.yml, as the easiest solution. Having this in dev-guide may help those building in Mac where many of the dependencies may be on a different path. (This seem to be applicable for linux builds as well but usually not required as installations usually made in standard search paths).

apriltuesday (approved):

nitin-ebi (approved):

PR #241, opened 2024-04-23
EVA-3281 - Allow synonyms to be the same value if they are on the same line

tcezard:
To trigger this error the assembly report had to contain a line with the same name be used multiple times and the same name be used in the fasta file.
The new integration test was able to reproduce [the error](https://github.com/EBIvariation/vcf-validator/actions/runs/8802923766/job/24159819760#step:4:722) and the subsequent commit fix the code


test/assembly_report/assembly_report_test.cpp
```diff
@@ -42,6 +42,14 @@ namespace ebi
             CHECK_NOTHROW(synonyms_map.parse_assembly_report(input));
         }
 
+        SECTION("Parsing assembly report with same value in one line")
+        {
+        ebi::assembly_report::SynonymsMap synonyms_map;
+        std::string assembly_report_path = "test/input_files/assembly_report/full_assembly_report/passed_with_same_synonym_on_same_line.txt";
+        std::ifstream input{assembly_report_path};
+        CHECK_NOTHROW(synonyms_map.parse_assembly_report(input));
+        }
```
apriltuesday, line 51:
```suggestion
        SECTION("Parsing assembly report with same value in one line")
        {
            ebi::assembly_report::SynonymsMap synonyms_map;
            std::string assembly_report_path = "test/input_files/assembly_report/full_assembly_report/passed_with_same_synonym_on_same_line.txt";
            std::ifstream input{assembly_report_path};
            CHECK_NOTHROW(synonyms_map.parse_assembly_report(input));
        }
```

apriltuesday (approved):

nitin-ebi (approved):

PR #242, opened 2024-04-24
GA4GHTT-261 - 44 1

vasudeva8:
-To share the changes made so far-
4.4 updated with changes for CN, SVCLAIM, SVLEN.
Newly added test files (*CNV*.vcf) are processed as required.
test_validator_v44 runs fine, test_validation_suit fails and needs further update.
Test code for changes to be updated.


src/vcf/record.cpp
```diff
@@ -704,7 +776,7 @@ namespace ebi
         }
         if(!values.empty()) {
             if (values.front() == MISSING_VALUE) { return; } // No need to check missing data
-        }
+        } //TODO, if the 1st one is . then check stops; is that correct? svclaim=.,DJ worked!
```
tcezard, line 779: This is not correct indeed: 
```
SVCLAIM=.
```
Is correct because missing value applies to all value
```
SVCLAIM=.,DJ
```
Needs to be checked for cardinality

vasudeva8:
taken care in PR #246. Closing this one.

PR #246, opened 2024-05-13
GA4GHTT-261 - changes for CN, SVLEN, SVCLAIM

vasudeva8:
4.4 updated, CN, SVLEN type change, SVCLAIM added.
Mandatory field check added, to deal with specific structural variants
BND removed from valid allele set
Tests and test files updated.

Assembler, debuggulator yet to update.
(The draft PR made earlier left as it is for reference as more validations are made)


vasudeva8:
A few of the 'should' cases were considered as 'must' and generated errors. They are now moved as optional checks to make warnings instead of errors.

src/vcf/record.cpp
```diff
@@ -306,6 +308,33 @@ namespace ebi
         }
     }
 
+    void Record::check_info_have_mandatory() const
+    {
+        static boost::regex deldup_regex("(<(DUP|DEL)(:[^>]+)*>)+");
+
+        if (source->version > Version::v43) {              //not applicable for 4.1/2/3
+            for (size_t i = 0; i < alternate_alleles.size(); ++i ) {
+                //SVLEN must for symbolic SV
```
apriltuesday, line 317:
```suggestion
                //SVLEN must be present for symbolic SV
```

apriltuesday (approved):
Looks good to me

tcezard (approved):

PR #247, opened 2024-06-07
update to generate mac arm binary

vasudeva8:
build.yml updated and added another job with macOS-14, which is arm based, to generate arm binaries.


tcezard (approved):
This looks good.
However there seem to be another dependency for linux that broke.
We can merge this branch for now but should look into what happened in the linux builds  separately.

PR #248, opened 2024-06-14
EVA-3586 - Allow colon in chromosome names

tcezard:
Simply allows colon in chromosome names


inc/vcf/error-odb.hpp
```diff
@@ -1052,7 +1052,7 @@ namespace odb
     typedef
     sqlite::query_column<
       sqlite::value_traits<
-        ::size_t,
+        long unsigned int,
```
tcezard, line 1055: I think these change have been introduce by ragel transpiler

src/vcf/vcf_v42.ragel
```diff
@@ -57,8 +57,8 @@
     #################################################
 
 
-    ## A chromosome must be a string with no white-spaces or colons, and may be surronded by < > symbols (for contigs)
-    chrom_basic     = alnum ( alnum | punct - (':' | '<' | '>' | ',') )* ;
+    ## A chromosome must be a string with no white-spaces or colons, and may be surrounded by < > symbols (for contigs)
```
apriltuesday, line 60:
```suggestion
    ## A chromosome must be a string with no white-spaces, and may be surrounded by < > symbols (for contigs)
```
and same for the other ragel files

apriltuesday (approved):
I have to plead ignorance on the `size_t` to `long unsigned int` change... but otherwise this looks fine.

PR #252, opened 2024-07-10
GA4GHTT-276 - minor update

vasudeva8:
update to remove a TODO.
  MISSING_VALUE can be a valid value for some, like SVCLAIM. So size is checked to see whether value is only '.' or more are present. If more are present, then cardinality is checked.

added comment on ragel file, showing genotype validation doesn't occur in ragel space as sample value definition is a super set of genotype definition. This necessitates header regeneration as well.


src/vcf/record.cpp
```diff
@@ -679,6 +673,60 @@ namespace ebi
...
             }
+        } else if (field_key == MEINFO) {
+            if (source->version < Version::v44) {   //not applicable for anything < v4.4
+                return;
+            }
+            if (values.size() != 4 * alternate_alleles.size()) {            //MEINFO must be 4 * ALT allele count
+                std::stringstream message;
+                message << "INFO " << MEINFO << " for record at " << line << " must have " << 4 * alternate_alleles.size() << " value(s)";
+                throw new InfoBodyError{line, message.str(), "Found " + std::to_string(values.size()) + " value(s)"};
+            }
```
tcezard, line 684: I feel like this should be refactored as a function to be used when cardinality is X time ALT.

tcezard (approved):

PR #257, opened 2024-09-16
GA4GHTT-303 - odb dependency removed, debugulator removed from build

vasudeva8:
ODB dependency removed,
Debugulator removed from build as it depends on odb.


vasudeva8:
Removed a few other debugulator related files as suggested.

tcezard (approved):

inc/vcf/error.hpp
```diff
@@ -128,7 +127,6 @@ namespace ebi
         Severity severity;
 
       private:
-        friend class odb::access;
```
tcezard, line 131: What is the purpose of the `#pragma db object polymorphic` above ? 
Can it be removed as well ? 
I think this (and the others) is the reason for the [warning suppression in CMakeLists.txt](https://github.com/EBIvariation/vcf-validator/blob/5084ea3181a2737ae6e37574a86f65b79dfd87eb/CMakeLists.txt#L100)

vasudeva8: will remove as a different PR

PR #266, opened 2024-10-17
fixes build issue in aarch64 observed in some builds

vasudeva8:
observed a build error during check of aarch64 linux on a vm.
from error, it has char as unsigned data type by default and hence couldn't hold -117 in the data.
explicitly made the type as signed char that it builds fine.


tcezard:
This was flagged in [bioconda](https://github.com/bioconda/bioconda-recipes/pull/51425) as well

tcezard (approved):

